PR #427, opened 2014-11-05
Ensure up-to-date snapshot is taken with the latest dictionary

charleso:
This started as something else, putting the check in `SquashJob.squashFromSnapshotWith`, but slowly moved up to where it is. I _think_ making the dictionary check in `SnapshotManifest` (or at least `Snapshots`) makes more sense.


markhibberd:
@charleso Is this irrelevant now, can you do a quick review of the code to make sure it is doing the right thing, but everything should be coming from the current commit I think, so should be all good.

charleso:
@markhibberd The current code looks correct. Might be good to ensure we only read the latest commit once, it's happening in `Snapshots.takeSnapshot` and `SquashJob.squashFromSnapshotWith`.

PR #483, opened 2014-11-21
Replace some dateutils, remove to seconds because it's unused

FrontierOtr:
This should give a much faster toDays and a more predictable minus days...

Also ditches toSeconds


markhibberd:
@FrontierOtr If you get bored at some point, can you have another look at this, just @charleso comments and passing tests and should be good.

PR #525, opened 2014-12-17
Add inverse reducer

FrontierOtr:


FrontierOtr:
This was folded into latestn

PR #532, opened 2014-12-18
Topic/latest nreducer

FrontierOtr:
WIP


ivory-core/src/main/scala/com/ambiata/ivory/core/Expression.scala
```diff
@@ -181,6 +207,63 @@ object Expression {
...
+   * will look like and they are currently represented below as [[StructEncoding]], but
+   * the keys are not known until later.
+   */
+  def expressionEncoding(expression: Expression, source: Encoding): Encoding = {
+    def getExpressionEncoding(exp: SubExpression, enc: Encoding): Encoding = exp match {
+      case Latest              => enc
+      case LatestN(_)          => enc match {
+                                    case sub: SubEncoding  => ListEncoding(sub)
+                                    // Nesting is forbidden, this is just to make the match exhaustive
+                                    case _  : ListEncoding => enc
```
FrontierOtr: Should latest N on a list just concatenate the entries?

ivory-core/src/main/scala/com/ambiata/ivory/core/Expression.scala
```diff
@@ -144,6 +152,14 @@ object Expression {
         case (Min | Max | Mean | Gradient | StandardDeviation) => ok
         case _  => "Non-supported interval sub expression".left
       }
+      case Inverse(other)                => {
```
charleso: Don't need the brace /supernit

ivory-core/src/main/scala/com/ambiata/ivory/core/Expression.scala
charleso: > Should latest N on a list just concatenate the entries?

I don't think that makes much sense does it? If/when we support recursive thrift types that would be one way to solve it properly.

ivory-core/src/test/scala/com/ambiata/ivory/core/gen/GenDictionary.scala
```diff
@@ -75,7 +77,11 @@ object GenDictionary {
     Gen.oneOf(fallback, cd.encoding match {
       case StructEncoding(values) =>
         val subexpGen = Gen.oneOf(values.toList).flatMap {
-          case (name, sve) => subExpression(sve.encoding).map(se => StructExpression(name, se))
+          case (name, sve) => Gen.oneOf(subExpression(sve.encoding).map(se => StructExpression(name, se)),
+                  sve.encoding match {
+                    case IntEncoding | LongEncoding | DoubleEncoding => numericSubs.map(se => Inverse(StructExpression(name, se)))
```
charleso: Wondering if we should add a `Encoding.isNumeric` combinator?

ivory-operation/src/test/scala/com/ambiata/ivory/operation/extraction/reduction/InverseReducerSpec.scala
```diff
@@ -0,0 +1,44 @@
...
+      value
+    }
+  }
+  class DummyIntReducer(ret: Int) extends Reduction {
+    val value = new ThriftFactValue()
+    def clear(): Unit = ()
+    def update(f: Fact): Unit = ()
+    def skip(f: Fact, reason: String): Unit = ()
+    def save: ThriftFactValue = {
+      value.setI(ret)
```
charleso: You could actually take `ReductionValueTo` in the constuctor here to remove the duplication if you wanted

ivory-operation/src/test/scala/com/ambiata/ivory/operation/extraction/reduction/LatestNReducerSpec.scala
```diff
@@ -0,0 +1,42 @@
...
+import scala.collection.JavaConverters._
+
+class LatestNReducerSpec extends Specification with ScalaCheck { def is = s2"""
+  LatestN reducer works with arbitrary facts       $latestN
+  LatestNStruct reducer works with arbitrary facts $latestNStruct
+
+"""
+
+  def latestN = prop((x: (Int, NonEmptyList[Fact])) => {
+    val num = Math.abs(x._1 % 7) + 1
```
charleso: Why `% 7`?

ivory-operation/src/test/scala/com/ambiata/ivory/operation/extraction/reduction/LatestNReducerSpec.scala
```diff
@@ -0,0 +1,42 @@
...
+    val xs = x._2.list.filterNot(_.toThrift.getValue.isSetLst).filterNot(_.toThrift.getValue.isSetT)
+    val r = new LatestNReducer(num)
+
+    xs.foreach(r.update)
```
charleso: `ReducerUtil.updateAll`

ivory-operation/src/test/scala/com/ambiata/ivory/operation/extraction/reduction/LatestNReducerSpec.scala
```diff
@@ -0,0 +1,42 @@
...
+
+    r.save.getLst.getL.asScala.toList ==== xs.reverse.take(num).map(_.toThrift.getValue).map { 
+      case tv if tv.isSetStructSparse => ThriftFactListValue.s(tv.getStructSparse)
+      case tv if tv.isSetD => ThriftFactListValue.p(ThriftFactPrimitiveValue.d(tv.getD))
+      case tv if tv.isSetS => ThriftFactListValue.p(ThriftFactPrimitiveValue.s(tv.getS))
+      case tv if tv.isSetI => ThriftFactListValue.p(ThriftFactPrimitiveValue.i(tv.getI))
+      case tv if tv.isSetL => ThriftFactListValue.p(ThriftFactPrimitiveValue.l(tv.getL))
+      case tv if tv.isSetB => ThriftFactListValue.p(ThriftFactPrimitiveValue.b(tv.getB))
+      case tv if tv.isSetDate => ThriftFactListValue.p(ThriftFactPrimitiveValue.date(tv.getDate))
```
charleso: Is it easier to manipulate the values at the `Fact/Value` level rather than having to do this?

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/reduction/InverseReducer.scala
```diff
@@ -0,0 +1,38 @@
+package com.ambiata.ivory.operation.extraction.reduction
+
+import com.ambiata.ivory.core.thrift._
+import com.ambiata.ivory.core.Fact
+import com.ambiata.ivory.core.Crash
```
charleso: `core.{Crash, Fact}` /supernit

ivory-operation/src/test/scala/com/ambiata/ivory/operation/extraction/reduction/LatestNReducerSpec.scala
FrontierOtr: Wanted a length which was arbitrary, but between 1 and a relatively small number (like 7) so a many meg piece of memory wouldn't be allocated and slow down the spec.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/reduction/LatestNReducer.scala
```diff
@@ -0,0 +1,82 @@
...
+    }
+    xs = xs.reverse
+    ThriftFactValue.lst(new ThriftFactList(xs.map {
+      case tv if tv.isSetStructSparse => ThriftFactListValue.s(tv.getStructSparse)
+      case tv if tv.isSetD => ThriftFactListValue.p(ThriftFactPrimitiveValue.d(tv.getD))
+      case tv if tv.isSetS => ThriftFactListValue.p(ThriftFactPrimitiveValue.s(tv.getS))
+      case tv if tv.isSetI => ThriftFactListValue.p(ThriftFactPrimitiveValue.i(tv.getI))
+      case tv if tv.isSetL => ThriftFactListValue.p(ThriftFactPrimitiveValue.l(tv.getL))
+      case tv if tv.isSetB => ThriftFactListValue.p(ThriftFactPrimitiveValue.b(tv.getB))
+      case tv if tv.isSetDate => ThriftFactListValue.p(ThriftFactPrimitiveValue.date(tv.getDate))
```
charleso: There should probably be at least two separate functions `toPrimitive: ThriftFactValue -> Option[ThriftFactPrimitiveValue]` and `fromPrimitive: ThriftFactPrimitiveValue -> ThriftFactValue` in `core` that we can write a bidirection property for.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/reduction/LatestNReducer.scala
```diff
@@ -0,0 +1,82 @@
...
+  def save: ThriftFactValue = {
+    var i = Math.min(count, n)
+    var xs: List[ThriftFactValue] = Nil
+    while (i > 0) {
+      hand = if (hand == 0) n-1 else (hand - 1) % n
+      i = i - 1
+      xs = clock(hand) :: xs
+    }
+    xs = xs.reverse
+    ThriftFactValue.lst(new ThriftFactList(xs.map {
```
charleso: I might be tempted to just create the raw `java.util.List` above and convert each value as you go to avoid creating multiple lists. /crazytown

ivory-operation/src/test/scala/com/ambiata/ivory/operation/extraction/reduction/InverseReducerSpec.scala
FrontierOtr: Indeed

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/reduction/LatestNReducer.scala
FrontierOtr: I was thinking here of only accepted structs and passing other values to the "pure" reducer below.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/reduction/LatestNReducer.scala
```diff
@@ -0,0 +1,82 @@
...
+import com.ambiata.ivory.core.Fact
+import com.ambiata.ivory.core.thrift._
+import com.ambiata.ivory.core.Crash
+import scala.collection.JavaConverters._
+
+class LatestNReducer(n: Int) extends Reduction {
+
+  val clock: Array[ThriftFactValue] = new Array[ThriftFactValue](n)
+  var hand: Int = 0
+  var count: Int = 0
```
charleso: Wondering whether you can/should use `ClockState` here and put some of the `update` and `save` logic in one place.

ivory-core/src/test/scala/com/ambiata/ivory/core/gen/GenDictionary.scala
```diff
@@ -83,13 +89,17 @@ object GenDictionary {
           ie <- values.find(v => v._1 != se && List(StringEncoding).contains(v._2.encoding)).map(ie => CountBySecondary(se, ie._1)) orElse
             values.find(v => List(IntEncoding, LongEncoding, DoubleEncoding).contains(v._2.encoding)).map(ie => SumBy(se, ie._1))
         } yield ie).cata(v => Gen.frequency(5 -> Gen.const(v), 5 -> subexpGen), subexpGen)
-      case p: PrimitiveEncoding   => subExpression(p).map(BasicExpression)
+      case p: PrimitiveEncoding   => Gen.oneOf(subExpression(p).map(BasicExpression),
+                  p match {
```
charleso: Identitation /supernit

ivory-core/src/main/scala/com/ambiata/ivory/core/Expression.scala
```diff
@@ -144,6 +152,14 @@ object Expression {
...
+        expressionEncoding(other, encoding) match {
+          case DoubleEncoding => ok
+          case LongEncoding   => ok
+          case IntEncoding    => ok
```
charleso: Here was the other place that could use `Encoding.isNumeric`

doc/design/feature-gen.md
```diff
@@ -21,6 +21,11 @@ less than or equal to `ref`);
 ### `latest`
 Return the most recent fact in `facts` with respect to `ref`.
 
+### `latestN`
+Return the most recent "n" facts if `facts` with respect to `ref`.
+Not available for "list" facts. Struct facts can optionally have an additional 
+`field` specified.
```
charleso: +1 for doc. :)

charleso:
Looks OK apart from a few comments. Definitely heading into some ugly territory with thrift values.

FWIW I still would have suggested that you keep the two PRs separate and just take the merge conflict hit pain (I know it sucks).

ivory-operation/src/test/scala/com/ambiata/ivory/operation/extraction/reduction/LatestNReducerSpec.scala
FrontierOtr: Messes up the ordering so gives different results.

FrontierOtr:
@markhibberd I think this is pretty much good to go. Would you mind giving it a quick look over please?

ivory-core/src/main/scala/com/ambiata/ivory/core/Encoding.scala
```diff
@@ -61,4 +61,12 @@ object Encoding {
       case _: StructEncoding    => false
       case _: ListEncoding      => false
     }
+
+  def isNumeric(enc: Encoding): Boolean =
+    enc match {
+      case IntEncoding       => true
+      case LongEncoding      => true
+      case DoubleEncoding    => true
+      case _                 => false
```
markhibberd: I know it is a pain but can you pattern match on everything, want to force a compile error that needs fixing if someone adds a case.

ivory-core/src/main/scala/com/ambiata/ivory/core/Expression.scala
markhibberd: I don't think I understand this, shouldn't we be making this an error for the moment?

FrontierOtr:
The failed test is erroneous (It was SquashSpec)

ivory-core/src/main/scala/com/ambiata/ivory/core/Encoding.scala
FrontierOtr: Ok

markhibberd:
Looks good. I am ok after pattern match change, and I saw later on the fail on List case, so I think that is ok, but just want to confirm that lastN of a List will fail for now right?

PR #551, opened 2015-01-06
Remove SampleFacts

charleso:
I ummed and ahhed about removing some of those specs. I'm not sure how much they were adding, given what else we already had. Now is probably a good time as any to replace them with something better if you think we need to.


ivory-operation/src/test/scala/com/ambiata/ivory/operation/extraction/output/SparseOutputSpec.scala
```diff
@@ -23,12 +22,11 @@ class SparseOutputSpec extends Specification with SampleFacts with ThrownExpecta
...
 """
 
-  def eav =
-    RepositoryBuilder.using(extractSparse(sampleFacts, sampleDictionary, TextEscaping.Delimited)) must beOkValue(
-      List("eid1|ns1|fid1|abc"
-         , "eid2|ns1|fid2|11"
-         , "eid3|ns2|fid3|true").sorted.mkString("\n") -> expectedDictionary
+  def eav = prop { (facts: FactsWithDictionary) =>
+    RepositoryBuilder.using(extractSparse(List(facts.facts), facts.dictionary, TextEscaping.Delimited)).map(_._1) must beOkLike(
+      (text: String) => seqToResult(text.split("\n").map(TextEscaping.split('|', _).length ==== 4))
```
charleso: Be nice to have something slightly better here. Could extract a function that's used in the reducer and ensure the output is the same?

charleso:
Note to self - need to fix `FactImporterSpec` as well

nhibberd:
Can remove the callbacks in the Spec's :smile:

PR #563, opened 2015-01-09
Correct Valid Date, add doc

FrontierOtr:
Woot!


charleso:
Worth adding a spec about min/max and valid date.

Otherwise :+1:

domdere:
I hate dealing with time

FrontierOtr:
Should I squash the two commits?

charleso:
Only if you want. It doesn't matter.

PR #567, opened 2015-01-12
Fix issues with Gen performance

charleso:


ivory-core/src/test/scala/com/ambiata/ivory/core/gen/GenDictionary.scala
```diff
@@ -174,7 +173,7 @@ object GenDictionary {
               case (name, StructEncodedValue(enc, _)) => filterExpression(enc).map(name ->)
             }).map(_.toList)
             // For 'and' we can only see each key once
-            cn     <- op.fold(Gen.const(1), Gen.choose(0, maxChildren))
+            cn     <- op.fold(Gen.const(1), Gen.choose(0, math.max(0, maxChildren)))
```
charleso: This was the subtle killer. When `maxChildren` got down to `-1` this would fail the entire `Gen`. The problem was that this was failing less than the default discard rate, which resulted in an increasing `size` that made the `Gen`s slower and slower.

See `org.scalacheck.Test.check().workerFun` for what I'm talking about.

ivory-core/src/test/scala/com/ambiata/ivory/core/gen/GenRepository.scala
```diff
@@ -21,8 +21,7 @@ object GenRepository {
   } yield Sized(v, s)
 
   def datasets: Gen[Datasets] = for {
-    n <- Gen.sized(z => Gen.choose(1, math.min(z, 20)))
```
charleso: Note to self (and others) - `size` will start off as `0`. This means that this will always be discarded. We should really stick to `GenPlus.listOfSized` where possible from now on.

ivory-core/src/test/scala/com/ambiata/ivory/core/gen/GenString.scala
```diff
@@ -13,7 +13,9 @@ object GenString {
     Gen.identifier
 
   def words: Gen[List[String]] =
-    Gen.listOf(word)
+    // FIX: This isn't really ideal, but be _very_ careful when changing the max size as this can result in
+    // _significantly_ slower property generation (see DatasetsSpec)
+    GenPlus.listOfSized(0, 5, word)
```
charleso: I'm happy to change/talk about this. Just for reference, this reduced the test times of `DatasetsSpec` (on my machine) from 126s to 25s. I'm not sure, at least for `Dictionary` whether this is worth the cost. We may want to have/call different versions for description and the tombstone list, which are the only consumers of this at the moment.

ivory-core/src/test/scala/com/ambiata/ivory/core/gen/GenString.scala
```diff
@@ -36,7 +38,7 @@ object GenString {
   def badName: Gen[String] =
     Gen.oneOf("", "_name", "name1/name2", "name㭊")
 
-  def colours: List[String] = List(
+  val colours: List[String] = List(
```
charleso: This was just for the hell of it.

ivory-core/src/test/scala/com/ambiata/ivory/core/gen/GenString.scala
markhibberd: I am ok with this,

PR #568, opened 2015-01-13
Introduce V3 to fix missing snapshot manifests

charleso:
/cc @raronson @nhibberd


ivory-operation/src/main/scala/com/ambiata/ivory/operation/update/UpdateV0.scala
```diff
@@ -73,10 +77,11 @@ object UpdateV0 {
 
     val key = Key(KeyName.unsafe(".metadata.json"))
 
-    def read(repository: Repository, s: Key): RIO[Option[SnapshotManifest]] = for {
-      exists <- repository.store.exists(s / SnapshotMetadataV1.key)
+    def read(repository: Repository, snapshotId: SnapshotId): RIO[Option[SnapshotManifest]] = for {
+      path   <- RIO.ok(Repository.snapshot(snapshotId) / key)
+      exists <- repository.store.exists(path)
       sm <- if (exists) for {
-        json <- repository.store.utf8.read(s / SnapshotMetadataV1.key)
```
charleso: My bad, this never worked. :(

ivory-operation/src/main/scala/com/ambiata/ivory/operation/update/UpdateV0.scala
```diff
@@ -19,17 +19,20 @@ object UpdateV0 {
...
+  def updateFactsets: RepositoryTIO[Unit] = RepositoryT.fromRIO(repository =>
+    Metadata.findFactsets(repository).flatMap(updateFactsetsIds(repository, _))
+  )
+
+  def updateFactsetsIds(repository: Repository, factsets: List[FactsetId]): RIO[Unit] = for {
     _ <- factsets.traverseU(f => for {
-      partitions <- Partitions.scrapeFromFactset(repository, f).map(p => (!p.isEmpty).option(p))
-      _    <- partitions.traverseU(ps => FactsetManifest.io(repository, f).write(FactsetManifest.create(f, FactsetFormat.V2, ps)))
+      partitions <- Partitions.scrapeFromFactset(repository, f)
+      _    <- FactsetManifest.io(repository, f).write(FactsetManifest.create(f, FactsetFormat.V2, partitions))
```
raronson: Should this be `V2` or `V3`?

charleso: > Should this be V2 or V3?

This is `FactsetFormat` which is different. :)

raronson: ah yep, sorry

raronson:
LGTM

ivory-operation/src/main/scala/com/ambiata/ivory/operation/update/UpdateV2.scala
```diff
@@ -0,0 +1,23 @@
...
+import scalaz._, Scalaz._
+
+object UpdateV2 {
+
+  def update: RepositoryTIO[Unit] =
+    for {
+      // The V0 code has been fixed, and we just need to re-run
+      _ <- UpdateV0.updateSnapshots
+      // Even empty factsets should have a manifest
+      _ <- updateEmptyFactsts
```
markhibberd: This should not happen, if you need to do this, it is covering a far worse bug in the factset read code. This needs to be changed back.

raronson: Im just curious - It is possible that empty factsets have been created and added to a store (it has happened in an existing repository). From my understanding, the previous updates did not add a manifest to empty factsets (even those within a store). Given this, how do you know if a factset should be skipped or hard fail in the factset read code? The contents of a factset dir might be accidentally removed and im not sure of a way to distinguish the difference between a valid empty factset (within a store) and one which has its contents removed.

PR #570, opened 2015-01-14
Ensure namespaceSizes only supports directories to avoid bad ingest errors

charleso:
This is to avoid confusing errors on CDH4 relating to executable permissions

### DIFF
--- a/ivory-storage/src/main/scala/com/ambiata/ivory/storage/fact/Namespaces.scala
+++ b/ivory-storage/src/main/scala/com/ambiata/ivory/storage/fact/Namespaces.scala
@@ -12,10 +12,13 @@ object Namespaces {
   /**
    * @return the list of namespaces for a given factset and their corresponding sizes
    */
-  def namespaceSizes(factsetPath: Path): Hdfs[List[(Namespace, BytesQuantity)]] =
-    Hdfs.childrenSizes(factsetPath)
+  def namespaceSizes(factsetPath: Path): Hdfs[List[(Namespace, BytesQuantity)]] = for {
+    dir <- Hdfs.isDirectory(factsetPath)
+    _   <- Hdfs.unless(dir)(Hdfs.fail(s"Invalid file $factsetPath for namespaces - must be a directory"))
+    ns  <- Hdfs.childrenSizes(factsetPath)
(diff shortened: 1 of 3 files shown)


ivory-storage/src/main/scala/com/ambiata/ivory/storage/fact/Namespaces.scala
```diff
@@ -12,10 +12,13 @@ object Namespaces {
   /**
    * @return the list of namespaces for a given factset and their corresponding sizes
    */
-  def namespaceSizes(factsetPath: Path): Hdfs[List[(Namespace, BytesQuantity)]] =
-    Hdfs.childrenSizes(factsetPath)
+  def namespaceSizes(factsetPath: Path): Hdfs[List[(Namespace, BytesQuantity)]] = for {
+    dir <- Hdfs.isDirectory(factsetPath)
+    _   <- Hdfs.unless(dir)(Hdfs.fail(s"Invalid file $factsetPath for namespaces - must be a directory"))
```
markhibberd: This should probably of failed before here in the `cli` right, this is assuming no namespace has been specified etc..?

markhibberd:
LGTM, maybe raise an issue to make sure validation is happening in cli apps on arguments as well.

markhibberd:
Oops, sorry.

ivory-storage/src/main/scala/com/ambiata/ivory/storage/fact/Namespaces.scala
charleso: > This should probably of failed before here in the cli right, this is assuming no namespace has been specified etc..?

charleso: > This should probably of failed before here in the cli right, this is assuming no namespace has been specified etc..?

I've added the validation to the cli as well. I'm assuming you don't want to remove this check from here? Otherwise the behaviour can be a little strange/dangerous for consumers.

markhibberd: This isn't general code, and probably shouldn't be masquerading as such, it is _very_ specific to ingest of unknown data in one of the specific ingest formats. I don't really mind too much if the check stays or goes, but I think a bit of code librarian action would be required to have this code make a little more sense and be less surprising.

PR #574, opened 2015-01-19
WIP: snapshot format v2

raronson:
@charleso @markhibberd This is still work in progress, but can you check if im on the right track. There are still a few tests which are failing, but its mostly done.


ivory-mr/src/main/scala/com/ambiata/ivory/mr/Mutator.scala
```diff
@@ -1,20 +1,19 @@
 package com.ambiata.ivory.mr
 
 import com.ambiata.poacher.mr.ThriftSerialiser
-import com.ambiata.ivory.core._
 import com.ambiata.ivory.core.thrift._
 import org.apache.hadoop.io.BytesWritable
 
-class ThriftByteMutator[T](implicit ev: T <:< ThriftLike) {
+class ThriftByteMutator {
```
charleso: This can be an `object` now

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/SnapshotJob.scala
```diff
@@ -249,25 +250,73 @@ class SnapshotIncrementalMapper extends CombinableMapper[NullWritable, BytesWrit
...
     okCounter = MrCounter("ivory", "snapshot.incr.ok", context)
     dropCounter = MrCounter("ivory", "drop", context)
   }
+}
+
+class SnapshotV1IncrementalMapper extends SnapshotIncrementalMapper[NullWritable] {
+  import SnapshotMapper._
+
+  /** Empty Fact, created once per mapper and mutated for each record */
+  val fact = new NamespacedThriftFact with NamespacedThriftFactDerived
```
charleso: FYI you can use `createMutableFact` which is on the ivory core package.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/SnapshotJob.scala
```diff
@@ -371,16 +427,19 @@ object SnapshotReducer {
...
+          emitter.emit(kout, vout)
           count += 1
         }
         datetime = fact.datetime
         // Store the current fact, which may or may not be emitted depending on the next fact
-        mutator.pipe(next, out)
+        kout.set(fact.date.int)
+        mutator.mutate(fact.toThrift, vout)
+        // emit to namespaced subdirs
+        emitter.path = "snapshot" + "/" + fact.namespace.name + "/part"
```
charleso: This will need to be cached. You should be able to build up an array for all the namespaces up front in the `setup` method.

ivory-mr/src/main/scala/com/ambiata/ivory/mr/Emitter.scala
```diff
@@ -2,12 +2,18 @@ package com.ambiata.ivory.mr
...
 import org.apache.hadoop.io.Writable
 import org.apache.hadoop.mapreduce.TaskInputOutputContext
+import org.apache.hadoop.mapreduce.lib.output.MultipleOutputs
 
 /** Abstraction over emitting key/value pairs in an mr job */
 trait Emitter[A, B] {
   def emit(key: A, value: B): Unit
 }
 
+trait MultiEmitter[A, B] extends Emitter[A, B] {
```
charleso: This doesn't need to extend `Emitter` does it? And then the `emit` function can take the `name` and `path`.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/squash/SquashMapReduce.scala
```diff
@@ -3,59 +3,92 @@ package com.ambiata.ivory.operation.extraction.squash
...
+
+class SquashV2Mapper extends SquashMapper[IntWritable] {
+  import SnapshotMapper._
+
+  val tfact = new ThriftFact
+  var namespace: String = null
+
+  override def setup(context: MapperContext[IntWritable]): Unit = {
+    super.setup(context)
+    namespace = Namespace.nameFromStringDisjunction(MrContext.getSplitPath(context.getInputSplit).getParent.getName) match {
```
charleso: Is it worth extracting a shared function for this somewhere?

ivory-operation/src/test/scala/com/ambiata/ivory/operation/extraction/ChordIncrementalMapperSpec.scala
```diff
@@ -9,25 +9,40 @@ import org.specs2._
...
-  def totals = prop((context: ChordMapperSpecContext, priority: Priority) => {
-    context.all.foreach(map(_, context, priority))
+  def counters2 = prop((context: ChordMapperSpecContext, priority: Priority) => {
+    context.all.foreach(mapV2(_, context, priority))
+    (context.ok.counter + context.skip.counter + context.drop.counter) ==== context.all.size
+  })
+
+  def totals1 = prop((context: ChordMapperSpecContext, priority: Priority) => {
+    context.all.foreach(mapV1(_, context, priority))
+    context.ok.counter ==== context.facts.size and
```
charleso: We've found that doing `(a, b, c) ==== (x, y, z)` can be more helpful than `a ==== x and b ==== y ...` /protip

project/build.scala
```diff
@@ -226,6 +226,9 @@ object build extends Build {
     else
       Seq()
 
+  // Scoobi sometimes swolloes exceptions, so if you want to see them do the following:
```
charleso: s/swolloes/swallows/ :)

ivory-core/src/test/scala/com/ambiata/ivory/core/gen/GenRepository.scala
```diff
@@ -15,6 +15,16 @@ object GenRepository {
...
     size.map(Bytes.apply)
 
+  def sizedNamespace: Gen[Sized[Namespace]] = for {
+    ns <- GenString.namespace
+    bs <- bytes
+  } yield Sized(ns, bs)
+
+  def namespaceBytes: Gen[List[Sized[Namespace]]] = for {
+    n   <- Gen.sized(z => Gen.choose(1, math.min(z, 5)))
+    nss <- Gen.listOfN(n, sizedNamespace)
```
charleso: `GenPlus.listOfSized(1, 5, sizedNamespace)`

The thing to be careful of is that `sized` can actually give you a 0. That and the combinator is nicer to read. :)

ivory-core/src/test/scala/com/ambiata/ivory/core/SnapshotSpec.scala
```diff
@@ -5,15 +5,24 @@ import com.ambiata.ivory.core.arbitraries.Arbitraries._
...
 
 """
+
+  def metadata = prop((s: Snapshot) =>
+    s.toMetadata ==== SnapshotMetadata(s.id, s.date, s.store.id, s.dictionary.map(_.id)))
+
+  def location = prop((snapshot: Snapshot) =>
+    snapshot.location ==== (snapshot.info match {
+      case SnapshotInfoV1(b)  => List(Repository.snapshot(snapshot.id))
+      case SnapshotInfoV2(bs) => bs.map(Repository.snapshot(snapshot.id) / _.value.asKeyName)
```
charleso: Hmm. Not sure there's much point to this test, it's the same thing as the `location` function isn't it? I don't have great ideas. You could check the size is > 0, or that it always starts with `Repository.snapshot(id)`?

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/ChordJob.scala
```diff
@@ -257,19 +259,68 @@ class ChordIncrementalMapper extends CombinableMapper[NullWritable, BytesWritabl
...
+  def mapV2(tfact: ThriftFact, key: IntWritable, value: BytesWritable, priority: Priority, kout: BytesWritable, vout: BytesWritable,
+            emitter: Emitter[BytesWritable, BytesWritable], okCounter: Counter, skipCounter: Counter, dropCounter: Counter,
+            serializer: ThriftSerialiser, featureIdLookup: FeatureIdLookup, entities: Entities, namespace: String) {
+    serializer.fromBytesViewUnsafe(tfact, value.getBytes, 0, value.getLength)
+    val date = Date.unsafeFromInt(key.get)
+    val fact = FatThriftFact(namespace, date, tfact)
+    val bytes = serializer.toBytes(fact)
+
+    // WARNING reusing input value BytesWritable
+    value.set(bytes, 0, bytes.length)
```
charleso: Is it worth writing a function to create a `Fact` from `ThriftFact` and `IntWritable` (this is the same as snapshot `mapV2`)?

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/squash/SquashJob.scala
```diff
@@ -52,11 +52,11 @@ object SquashJob {
    * 2. From a chord, where a single entity may have one or more dates. It will be important to pre-calculate a starting
    *    date for every possible entity date, and then look that up per entity on the reducer.
    */
-  def squash(repository: Repository, dictionary: Dictionary, input: ShadowOutputDataset, conf: SquashConfig,
+  def squash(repository: Repository, dictionary: Dictionary, inputs: List[Path], conf: SquashConfig,
              job: (Job, MrContext), cluster: Cluster): RIO[ShadowOutputDataset] = for {
     // This is about the best we can do at the moment, until we have more size information about each feature
```
charleso: Not for long. :)

ivory-mr/src/main/scala/com/ambiata/ivory/mr/Mutator.scala
raronson: I don't think it can because the `ThriftSerializer` member isn't thread safe

charleso: > I don't think it can because the ThriftSerializer member isn't thread safe

You're right, it definitely isn't. The implication of making it `object` would be to have to pass `ThriftSerializer` into the `from`/`to` and pass that around instead of `ThriftByteMutator`. /shrug

markhibberd: I will look at this tonight.

ivory-core/src/main/scala/com/ambiata/ivory/core/Snapshot.scala
```diff
@@ -1,12 +1,37 @@
...
+
+import scalaz._, Scalaz._
+
 case class Snapshot(
   id: SnapshotId
 , date: Date
 , store: FeatureStore
 , dictionary: Option[Identified[DictionaryId, Dictionary]]
-, bytes: Bytes
+, info: SnapshotInfo
```
markhibberd: This doesn't make sense to me. How the bytes are stored should be independent of the format.

ivory-core/src/main/scala/com/ambiata/ivory/core/Snapshot.scala
```diff
@@ -1,12 +1,37 @@
...
 ) {
   def toMetadata: SnapshotMetadata =
     SnapshotMetadata(id, date, store.id, dictionary.map(_.id))
+
+  def location: List[Key] = {
```
markhibberd: Why do you need this, this looks like it will lead to bugs, we shouldn't have anything to do with file paths, or keys on these data types.

ivory-core/src/main/scala/com/ambiata/ivory/core/Snapshot.scala
```diff
@@ -1,12 +1,37 @@
...
+    val base = Repository.snapshot(id)
+    info match {
+      case SnapshotInfoV1(_)     => List(base)
+      case SnapshotInfoV2(bytes) => bytes.map(s => base / s.value.asKeyName)
+    }
+  }
+}
+
+sealed trait SnapshotInfo {
```
markhibberd: Related to above, these should be two independent fields.

ivory-core/src/test/scala/com/ambiata/ivory/core/arbitraries/Arbitraries.scala
```diff
@@ -146,6 +146,9 @@ trait Arbitraries {
   implicit def SnapshotFormatArbitrary: Arbitrary[SnapshotFormat] =
     Arbitrary(GenVersion.snapshot)
 
+  implicit def SnapshotInfoArbitrary: Arbitrary[SnapshotInfo] =
```
markhibberd: Again, I don't think this is a meaningful type, I would really prefer it didn't exist.

ivory-core/src/test/scala/com/ambiata/ivory/core/gen/GenRepository.scala
```diff
@@ -15,6 +15,16 @@ object GenRepository {
   def bytes: Gen[Bytes] =
     size.map(Bytes.apply)
 
+  def sizedNamespace: Gen[Sized[Namespace]] = for {
+    ns <- GenString.namespace
+    bs <- bytes
+  } yield Sized(ns, bs)
```
markhibberd: Is there, can there be a generic sized gen?

ivory-mr/src/main/scala/com/ambiata/ivory/mr/Emitter.scala
```diff
@@ -2,12 +2,18 @@ package com.ambiata.ivory.mr
...
+  var name: String
+  var path: String
```
markhibberd: Why var's?

ivory-mr/src/main/scala/com/ambiata/ivory/mr/Emitter.scala
markhibberd: Why does this need a different abstraction then the code that already exists? There should only be one pattern required as far as I can see, can this just be done how factsets are done, or should factsets be changed to how this is done? Either way there shouldn't be two different patterns.

ivory-mr/src/main/scala/com/ambiata/ivory/mr/Mutator.scala
markhibberd: I agree, that would be much better to pass it as an argument and make this an object.

ivory-mr/src/test/scala/com/ambiata/ivory/mr/MockFactMutator.scala
```diff
@@ -21,10 +22,22 @@ object MockFactMutator {
         ()
       }
     }
-    val result = iterateFactsAsBytes(facts)(iter => f(iter, new FactByteMutator, emitter, Writables.bytesWritable(4096)))
+    val result = iterateFactsAsBytes(facts)(iter => f(iter, new ThriftByteMutator, emitter, Writables.bytesWritable(4096)))
     (outFacts.toList, result)
   }
 
+  def runV2Keep[A](facts: List[Fact])(f: (JIterator[BytesWritable], ThriftByteMutator, MultiEmitter[IntWritable, BytesWritable], IntWritable, BytesWritable) => A): (List[(String, Fact)], A) = {
```
markhibberd: Why is this just new code, shouldn't it just be a change to the existing code? As in, it strikes me that there should _only_ be the new way in the code base now.

ivory-core/src/test/scala/com/ambiata/ivory/core/gen/GenRepository.scala
charleso: > Is there, can there be a generic sized gen?

Good point, we/you have already created:

```
  implicit def SizedArbitrary[A: Arbitrary]: Arbitrary[Sized[A]] =
    Arbitrary(GenRepository.sized)
```

ivory-operation/src/main/scala/com/ambiata/ivory/operation/debug/DumpFactsJob.scala
```diff
@@ -102,6 +108,34 @@ class DumpFactsSnapshotMapper extends Mapper[NullWritable, BytesWritable, NullWr
   }
 }
 
+class DumpFactsSnapshotV1Mapper extends DumpFactsSnapshotMapper[NullWritable] {
```
markhibberd: We should be handling versioning consistently between factsets and snapshots. If this change is pushing out that we need different concrete classes for each version, then the same change needs to be made for factsets at the same time, I don't want more inconsistencies creeping into these jobs if we can help it.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/debug/DumpFactsJob.scala
```diff
@@ -40,9 +44,13 @@ object DumpFactsJob {
...
-        request.snapshots.foreach(id => {
-          val path = repository.toIvoryLocation(Repository.snapshot(id)).toHdfsPath
-          MultipleInputs.addInputPath(job, path, classOf[SequenceFileInputFormat[_, _]], classOf[DumpFactsSnapshotMapper])
+        ss.foreach(snapshot => {
+          val paths = snapshot.location.map(k => repository.toIvoryLocation(k).toHdfsPath)
+          val mapperClass = snapshot.info.format match {
+            case SnapshotFormat.V1 => classOf[DumpFactsSnapshotV1Mapper]
+            case SnapshotFormat.V2 => classOf[DumpFactsSnapshotV2Mapper]
+          }
+          paths.foreach(p => MultipleInputs.addInputPath(job, p, classOf[SequenceFileInputFormat[_, _]], mapperClass))
```
markhibberd: Should/Can this use IvoryInputs.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/debug/DumpFactsJob.scala
```diff
@@ -102,6 +108,34 @@ class DumpFactsSnapshotMapper extends Mapper[NullWritable, BytesWritable, NullWr
...
+
+  override def setup(context: Mapper[IntWritable, BytesWritable, NullWritable, Text]#Context): Unit = {
+    super.setup(context)
+    namespace = Namespace.nameFromStringDisjunction(MrContext.getSplitPath(context.getInputSplit).getParent.getName) match {
+      case scalaz.\/-(n) => n.name
+      case scalaz.-\/(e) => Crash.error(Crash.DataIntegrity, s"Can not parse snapshot namespace from path ${e}")
+    }
+  }
+
+  override def map(key: IntWritable, value: BytesWritable, context: Mapper[IntWritable, BytesWritable, NullWritable, Text]#Context): Unit = {
```
markhibberd: Should we consider BytesWritable as the key, as a step towards variable sized partition key?

ivory-core/src/main/scala/com/ambiata/ivory/core/Snapshot.scala
raronson: I was stuck on this actually because i had the following before, but found it a bit clumsy when trying to create a list of paths out of `format` and `bytes` (i would match on `format` then expect `bytes` to be either `left` or `right` depending on the `format`). How should i approach this?

```
case class Snapshot(
id: SnapshotId
, date: Date
, store: FeatureStore
, dictionary: Option[Identified[DictionaryId, Dictionary]]
, bytes: Bytes \/ List[Sized[Namespace]]
, format: SnapshotFormat
)
```

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/SnapshotJob.scala
```diff
@@ -109,11 +112,12 @@ object SnapshotJob {
     val FactsetVersionLookup = ThriftCache.Key("factset-version-lookup")
     val WindowLookup = ThriftCache.Key("factset-window-lookup")
     val FeatureIsSetLookup = ThriftCache.Key("feature-is-set-lookup")
+    val Out = "snap"
```
markhibberd: Inconsistency with above?

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/squash/SquashJob.scala
```diff
@@ -64,31 +64,47 @@ object SquashJob {
...
+      println(s"Input path: $input")
+      MultipleInputs.addInputPath(job, input, classOf[SequenceFileInputFormat[_, _]], mapperClass)
+    })
+
     job.getConfiguration.set(SnapshotJob.Keys.SnapshotDate, date.int.toString)
     (job, ctx)
   }
 
-  def initChordJob(conf: Configuration, chord: Entities): RIO[(Job, MrContext)] = RIO.safe {
+  def initChordJob(conf: Configuration, chord: Entities, input: Path): RIO[(Job, MrContext)] = RIO.safe {
```
markhibberd: This shouldn't be taking a Path as an argument. Need to pick or build an appropriate data type to represent what is being passed along.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/ChordJob.scala
```diff
@@ -257,19 +259,68 @@ class ChordIncrementalMapper extends CombinableMapper[NullWritable, BytesWritabl
...
+  override def map(key: IntWritable, value: BytesWritable, context: MapperContext[IntWritable]): Unit = {
+    emitter.context = context
+    ChordIncrementalMapper.mapV2(tfact, key, value, Priority.Max, kout, vout, emitter, okCounter, skipCounter, dropCounter, serializer, featureIdLookup, entities, namespace)
   }
 }
 
 object ChordIncrementalMapper {
 
-  def map(fact: NamespacedThriftFact with NamespacedThriftFactDerived, bytes: BytesWritable, priority: Priority,
+  def mapV1(fact: NamespacedThriftFact with NamespacedThriftFactDerived, bytes: BytesWritable, priority: Priority,
```
markhibberd: This looks a bit odd, can these just be inlined? And is there a better, smaller spot to fork on the different ways to build a fact. I would expect a single piece of code to go from input K, V => Fact and that be the only difference.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/IvoryInputs.scala
```diff
@@ -30,7 +30,7 @@ object IvoryInputs {
       case FactsetDataset(factset) =>
         Nil
       case SnapshotDataset(snapshot) =>
-        List(repository.toIvoryLocation(Repository.snapshot(snapshot.id)).toHdfsPath)
+        snapshot.location.map(k => repository.toIvoryLocation(k).toHdfsPath)
```
markhibberd: Related to comment in Snapshot.scala, it shouldn't know about Locations or Keys.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/SnapshotJob.scala
```diff
@@ -51,12 +50,16 @@ object SnapshotJob {
     job.setOutputValueClass(classOf[BytesWritable])
 
     // input
-    IvoryInputs.configure(ctx, job, repository, plan.datasets, classOf[SnapshotFactsetMapper], classOf[SnapshotIncrementalMapper])
+    val incrementalMapper = plan.snapshot.map(_.info.format match {
+      case SnapshotFormat.V1 => classOf[SnapshotV1IncrementalMapper]
+      case SnapshotFormat.V2 => classOf[SnapshotV2IncrementalMapper]
+    }).getOrElse(classOf[SnapshotV1IncrementalMapper])
```
markhibberd: Can this be pulled out into somewhere common in `ivory.storage.fact`

markhibberd: It looks weird to have to supply a default here, maybe IvoryInputs should change, but if not, the default should be V2 I guess?

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/SnapshotJob.scala
```diff
@@ -249,25 +250,73 @@ class SnapshotIncrementalMapper extends CombinableMapper[NullWritable, BytesWrit
...
   }
 }
 
 object SnapshotIncrementalMapper {
-  def map(fact: NamespacedThriftFact with NamespacedThriftFactDerived, bytes: BytesWritable, priority: Priority,
+  def mapV1(fact: NamespacedThriftFact with NamespacedThriftFactDerived, bytes: BytesWritable, priority: Priority,
           kout: BytesWritable, vout: BytesWritable, emitter: Emitter[BytesWritable, BytesWritable], okCounter: Counter,
           dropCounter: Counter, serializer: ThriftSerialiser, featureIdLookup: FeatureIdLookup) {
     serializer.fromBytesViewUnsafe(fact, bytes.getBytes, 0, bytes.getLength)
+    emit(fact, bytes, priority, kout, vout, emitter, okCounter, dropCounter, serializer, featureIdLookup)
```
markhibberd: Same comment as chord, doesn't really feel like the right split.

project/build.scala
markhibberd: There is a different issue already raised for this, you don't have to disable forking, you have to set the properties in the right spot.

ivory-core/src/main/scala/com/ambiata/ivory/core/Snapshot.scala
markhibberd: format shouldn't have any affect on what you should expect, ignore format and match on bytes when you are trying to work out what to do. You are getting yourself into a hole how it is at the moment, because everything will break when you have to tick the format for some other reason and you will be in a world of pain.

markhibberd:
Battery going flat, but good progress. Main concerns are consistency between handling factset and snapshot version differences in MR jobs, want them to be really consistent.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/debug/DumpFactsJob.scala
raronson: Im pretty sure we need to create multiple concrete classes for the different snapshot versions as the key class is different between the two. Ill update the factset version code to be consistent.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/debug/DumpFactsJob.scala
raronson: I was in two minds about this and had a discussion with @charleso when i was originally trying to decide between `IntWritable` and `BytesWritable`. We weren't sure which one to choose, so i picked `IntWritable` because its easier to use (just call `get` on it), and if we have to create a new mapper class per version, we can change it to `BytesWritable` when we come to variable sized partition keys. Im happy to change it now though.

ivory-mr/src/main/scala/com/ambiata/ivory/mr/Emitter.scala
raronson: I think i need to go over this with you, im not sure how it should look

ivory-mr/src/test/scala/com/ambiata/ivory/mr/MockFactMutator.scala
raronson: This is new because chord tests still use the old `NamespacedThrift` format, but snapshot uses the new format. I will renamed the functions.

raronson:
This still isn't completed yet, but i have incorporated most of the feedback

ivory-storage/src/main/scala/com/ambiata/ivory/storage/lookup/ReducerLookups.scala
```diff
@@ -45,14 +45,11 @@ object ReducerLookups {
...
-  def index(dictionary: Dictionary): (NamespaceLookup, FeatureIdLookup) =
-    indexDefinitions(dictionary.definitions)
-
-  def indexDefinitions(definitions: List[Definition]): (NamespaceLookup, FeatureIdLookup) = {
+  def index(dictionary: Dictionary): (NamespaceLookup, FeatureIdLookup) = {
     val namespaces = new NamespaceLookup
     val features = new FeatureIdLookup
 
-    definitions.zipWithIndex.foreach { case (d, idx) =>
+    dictionary.byFeatureIndex.foreach { case (idx, d) =>
```
raronson: Made this consistent with `FeatureLookups`

ivory-core/src/test/scala/com/ambiata/ivory/core/gen/GenRepository.scala
raronson: Im not sure what to do here as i don't have `Arbitrary[Namespace]` in scope.

raronson:
@markhibberd @charleso This is now ready for another look when you have time

ivory-core/src/test/scala/com/ambiata/ivory/core/gen/GenRepository.scala
charleso: > Im not sure what to do here as i don't have Arbitrary[Namespace] in scope.

Didn't you decide to change `GenRepository.sized` to `def sized(g: Gen[A]): Gen[Sized[A]]` and you call it directly?

doc/user/api-compatibility.md
```diff
@@ -3,6 +3,10 @@ API Compatibility
 
 This document serves as a list of changes to the Ivory API, both in the file format and in the Scala API.
 
+## 1.0.0-*-xxxxxxx
+
+- printFacts in the `Ivory` api has been replaced with `dumpFactsToFile` and `dumpFactsToStdout`. 
+
```
markhibberd: Worth mentioning cli change as well.

ivory-core/src/main/scala/com/ambiata/ivory/core/Snapshot.scala
```diff
@@ -1,12 +1,21 @@
...
+, bytes: Bytes \/ List[Sized[Namespace]]
+, format: SnapshotFormat
 ) {
   def toMetadata: SnapshotMetadata =
     SnapshotMetadata(id, date, store.id, dictionary.map(_.id))
+
+  def totalBytes: Bytes =
+    bytes match {
+      case -\/(b)  => b
+      case \/-(bs) => bs.foldMap(_.bytes)
```
charleso: FWIW I prefer `fold` but that's just a personal thing.

markhibberd: Or in this case it is just `bytes.swap.getOrElse(_.foldMap(_,bytes))`.

ivory-core/src/main/scala/com/ambiata/ivory/core/package.scala
```diff
@@ -1,10 +1,13 @@
...
+import com.ambiata.ivory.core.thrift.{NamespacedThriftFact, ThriftFact}
 
 package object core {
   type MutableFact = NamespacedThriftFact with NamespacedThriftFactDerived
 
-  def createMutableFact: MutableFact =
-    new NamespacedThriftFact with NamespacedThriftFactDerived
+  def createMutableFact: MutableFact = {
+    val fact = new NamespacedThriftFact with NamespacedThriftFactDerived
+    fact.setFact(new ThriftFact)
```
markhibberd: What drove this change?

ivory-operation/src/main/scala/com/ambiata/ivory/operation/debug/DumpFactsJob.scala
```diff
@@ -75,56 +80,99 @@ object DumpFactsJob {
...
     val entities = DumpFactsJob.read(context.getConfiguration, DumpFactsJob.Keys.Entities).toSet
     val attributes = DumpFactsJob.read(context.getConfiguration, DumpFactsJob.Keys.Attributes).toSet
     mapper = DumpFactsMapper(entities, attributes, source)
-    fact = FatThriftFact(p.namespace.name, p.date, underlying)
+    setupSplitFormat(context, split)
   }
 
-  override def map(key: NullWritable, value: BytesWritable, context: Mapper[NullWritable, BytesWritable, NullWritable, Text]#Context): Unit = {
-    serializer.fromBytesViewUnsafe(underlying, value.getBytes, 0, value.getLength)
+  def setupSplitFormat(context: Mapper[K, BytesWritable, NullWritable, Text]#Context, split: InputSplit): Unit
```
charleso: Could this return `MrFactConverter` instead?

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/ChordJob.scala
```diff
@@ -138,43 +145,52 @@ class ChordFactsetMapper extends CombinableMapper[NullWritable, BytesWritable, B
...
+    okCounter = MrCounter("ivory", s"chord.v${format.toStringFormat}.ok", context)
+    skipCounter = MrCounter("ivory", s"chord.v${format.toStringFormat}.skip", context)
     dropCounter = MrCounter("ivory", "drop", context)
-    converter = factsetInfo.factConverter
+    partition = factsetInfo.partition
     priority = factsetInfo.priority
+    setupSplitFormat(context, split)
   }
 
+  def setupSplitFormat(context: MapperContext[K], split: InputSplit): Unit
```
charleso: Same here, return `MrFactConverter`?

ivory-core/src/test/scala/com/ambiata/ivory/core/arbitraries/RepositoryScenario.scala
```diff
@@ -109,21 +109,26 @@ object RepositoryScenario {
         chance    <- Gen.choose(1, 10).map(_ < 3)
         earliest  <- Gen.choose(1, day * span * 5)
         latest    <- Gen.choose(1, day * span * 5)
-        bytes     <- arbitrary[Bytes]
         format    <- arbitrary[FactsetFormat]
+        bytes     <- arbitrary[Bytes]
+        sformat   <- arbitrary[SnapshotFormat]
+        sbytes    <- sformat match {
```
markhibberd: I wouldn't do this switch, I would just generate one or the other, with a bias towards the list, i.e. `frequency(1 -> arb[Bytes].map(_.left), 9 -> arb[...].map(_.right)`

ivory-core/src/test/scala/com/ambiata/ivory/core/gen/GenRepository.scala
```diff
@@ -66,8 +74,12 @@ object GenRepository {
     x <- GenDate.date
     s <- store
     d <- GenDictionary.identified
-    b <- bytes
-  } yield Snapshot(i, x, s, d.some, b)
+    f <- GenVersion.snapshot
+    b <- f match {
```
markhibberd: Same as previous comment, seems to be a bit of dupe with the other gen code here?

ivory-mr/src/main/scala/com/ambiata/ivory/mr/Mutator.scala
```diff
@@ -1,20 +1,17 @@
...
-  val serializer = ThriftSerialiser()
-
-  def from(in: BytesWritable, thrift: T): Unit = {
+  def from[T](in: BytesWritable, thrift: T, serializer: ThriftSerialiser)(implicit ev: T <:< ThriftLike): Unit = {
     serializer.fromBytesViewUnsafe(thrift, in.getBytes, 0, in.getLength)
     ()
   }
 
-  def mutate(in: T, vout: BytesWritable): Unit = {
+  def mutate[T](in: T, vout: BytesWritable, serializer: ThriftSerialiser)(implicit ev: T <:< ThriftLike): Unit = {
```
markhibberd: Should this just be `T <: ThriftLike` and no implicit ev?

ivory-core/src/main/scala/com/ambiata/ivory/core/package.scala
raronson: Some of the implementations of `MrFactConverter` expects the given `MutableFact` to have a `ThriftFact`, which it mutates, but when `NamespacedThriftFact with NamespacedThriftFactDerived` is constructed, its `null` by default.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/debug/DumpFacts.scala
```diff
@@ -19,10 +31,98 @@ case class DumpFactsRequest(
...
+    repository match {
+      case hr @ HdfsRepository(_) => for {
+        paths <- IvoryInputs.factsetPaths(hr, factset).traverse(expandPath).map(_.flatten).run(hr.configuration)
+        ret   <- paths.traverse(path => factset.format match {
+          case FactsetFormat.V1 => printV1Factset(path, hr.configuration, mapper)
+          case FactsetFormat.V2 => printV2Factset(path, hr.configuration, mapper)
+        }).map(_.sequenceU.void)
+      } yield ret
+      case LocalRepository(_) => RIO.ok("DumpFacts does not support reading factsets from local repositories yet".left)
+      case S3Repository(_) => RIO.ok("DumpFacts does not support reading factsets from s3 repositories yet".left)
```
markhibberd: Was that the case before? I thought this was working? (could very well be mistaken though).

ivory-operation/src/main/scala/com/ambiata/ivory/operation/debug/DumpFacts.scala
```diff
@@ -19,10 +31,98 @@ case class DumpFactsRequest(
...
+  }
+
+  def printV1Factset(path: Path, config: Configuration, mapper: DumpFactsMapper): RIO[String \/ Unit] = {
+    Partition.parseDir(DirPath.unsafe(path.toString)).disjunction.traverse(partition => {
+      val converter = PartitionFactConverter(partition)
+      Print.printWith(path, config, NullWritable.get, Writables.bytesWritable(4096))(print(createMutableFact, converter, mapper) _)
+    })
+  }
+
+  def printV2Factset(path: Path, config: Configuration, mapper: DumpFactsMapper): RIO[String \/ Unit] = {
```
markhibberd: Dupe with above?

ivory-operation/src/main/scala/com/ambiata/ivory/operation/debug/DumpFactsJob.scala
```diff
@@ -75,56 +80,99 @@ object DumpFactsJob {
...
-class DumpFactsSnapshotMapper extends Mapper[NullWritable, BytesWritable, NullWritable, Text] {
+abstract class DumpFactsFactsetMapper[K <: Writable] extends CombinableMapper[K, BytesWritable, NullWritable, Text] {
   val serializer = ThriftSerialiser()
   val buffer = new StringBuilder(4096)
-  val fact = new NamespacedThriftFact with NamespacedThriftFactDerived
-  val key = NullWritable.get
-  val out = new Text
-  val missing = "NA"
-
+  val underlying = new ThriftFact
```
markhibberd: Is this used?

ivory-operation/src/main/scala/com/ambiata/ivory/operation/debug/DumpFacts.scala
raronson: Its just took `Path` objects before, straight from the cli.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/debug/DumpFactsJob.scala
```diff
@@ -75,56 +80,99 @@ object DumpFactsJob {
...
-      context.write(key, out)
+      vout.set(mapper.renderWith(fact, buffer))
+      context.write(kout, vout)
     }
   }
 }
 
-class DumpFactsFactsetMapper extends Mapper[NullWritable, BytesWritable, NullWritable, Text] {
+class DumpFactsV1FactsetMapper extends DumpFactsFactsetMapper[NullWritable] {
+  override def setupSplitFormat(context: Mapper[NullWritable, BytesWritable, NullWritable, Text]#Context, split: InputSplit): Unit = {
```
markhibberd: If this is the only likely difference, could this just return the value rather than mutating the super field?

raronson: Yep, @charleso also recommended it

ivory-operation/src/main/scala/com/ambiata/ivory/operation/debug/DumpFacts.scala
```diff
@@ -19,10 +31,98 @@ case class DumpFactsRequest(
...
+  } yield ret
+
+  def datasets(repository: Repository, request: DumpFactsRequest): RIO[String \/ Datasets] = for {
+    factsets  <- request.factsets.traverse(fid => Factsets.factset(repository, fid))
+    snapshots <- request.snapshots.traverse(sid => SnapshotStorage.byIdOrFail(repository, sid))
+    pdatasets  = Prioritized.fromList(factsets.map(Dataset.factset) ++ snapshots.map(Dataset.snapshot))
+    datasets   = pdatasets.cata(ds => Datasets(ds).right, "Too many factsets/snapshots!".left)
+  } yield datasets
+
+  def print[K <: Writable, V <: Writable](fact: MutableFact, converter: MrFactConverter[K, V], mapper: DumpFactsMapper)(key: K, value: V): IO[Unit] = {
```
markhibberd: Would some of this code be better off living with the other display/print code, seems tangental to the dump code?

ivory-mr/src/main/scala/com/ambiata/ivory/mr/Emitter.scala
```diff
@@ -2,16 +2,27 @@ package com.ambiata.ivory.mr
...
-trait Emitter[A, B] {
+trait ContextEmitter[A, B] {
   def emit(key: A, value: B): Unit
 }
 
-case class MrEmitter[IK <: Writable, IV <: Writable, OK <: Writable, OV <: Writable]() extends Emitter[OK, OV] {
-  var context: TaskInputOutputContext[IK, IV, OK, OV] = null
+trait OutputEmitter[A, B] {
+  def emitPath(key: A, value: B, path: String): Unit
+  def close(): Unit
```
charleso: Do we even need `close()` to be on the trait? I'm assuming/guessing that this will only be called on `MrOutputEmitter` anyway.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/ChordJob.scala
```diff
@@ -53,12 +52,19 @@ object ChordJob {
...
 
-    IvoryInputs.configure(ctx, job, repository, plan.datasets, classOf[ChordFactsetMapper], classOf[ChordIncrementalMapper])
+    // input
+    IvoryInputs.configure(ctx, job, repository, plan.datasets, {
+      case FactsetFormat.V1 => classOf[ChordV1FactsetMapper]
+      case FactsetFormat.V2 => classOf[ChordV2FactsetMapper]
+    }, {
+      case SnapshotFormat.V1 => classOf[ChordV1IncrementalMapper]
+      case SnapshotFormat.V2 => classOf[ChordV2IncrementalMapper]
+    })
```
markhibberd: This looks better :+1:

ivory-operation/src/main/scala/com/ambiata/ivory/operation/debug/DumpFactsJob.scala
raronson: nope, removing

ivory-mr/src/main/scala/com/ambiata/ivory/mr/Emitter.scala
```diff
@@ -2,16 +2,27 @@ package com.ambiata.ivory.mr
...
 /** Abstraction over emitting key/value pairs in an mr job */
-trait Emitter[A, B] {
+trait ContextEmitter[A, B] {
   def emit(key: A, value: B): Unit
 }
 
-case class MrEmitter[IK <: Writable, IV <: Writable, OK <: Writable, OV <: Writable]() extends Emitter[OK, OV] {
-  var context: TaskInputOutputContext[IK, IV, OK, OV] = null
+trait OutputEmitter[A, B] {
+  def emitPath(key: A, value: B, path: String): Unit
```
markhibberd: Pedantic, but I think this should be called `emit` to be consistent with other one now that they are going to be on separate traits.

markhibberd:
Looks good, but I think you missed a bit ;) Don't all the cli tests use cat facts?

ivory-mr/src/main/scala/com/ambiata/ivory/mr/Emitter.scala
charleso: > Pedantic, but I think this should be called emit to be consistent with other one now that they are going to be on separate traits.

Just discussed with @raronson IRL - the other alternative is we put `path` on the constructor or `MrOutputEmitter` and create it each time, and then we only need the single `ContextEmitter`. Would that be better? Or do we prefer the split?

markhibberd: If it works out that is cool, I thought it was tried and failed, because the path comes from values in certain situations, but if that is not the case then single interface is a lot nicer.

charleso: > If it works out that is cool, I thought it was tried and failed, because the path comes from values in certain situations, but if that is not the case then single interface is a lot nicer.

I believe it's possible, the only "downside" is the extra allocation per map/reduce if the path needs to be dynamic (but I don't see that as a problem if we're already doing string allocations)

ivory-operation/src/main/scala/com/ambiata/ivory/operation/debug/DumpFacts.scala
```diff
@@ -19,10 +31,98 @@ case class DumpFactsRequest(
...
+
+  def printV1Snapshot(path: Path, config: Configuration, mapper: DumpFactsMapper): RIO[String \/ Unit] = {
+    val converter = MutableFactConverter()
+    Print.printWith(path, config, NullWritable.get, Writables.bytesWritable(4096))(print(createMutableFact, converter, mapper) _).map(_.right)
+  }
+
+  def printV2Snapshot(path: Path, config: Configuration, mapper: DumpFactsMapper): RIO[String \/ Unit] = {
+    Namespace.nameFromStringDisjunction(path.getName).traverse(namespace => {
+      val converter = NamespaceDateFactConverter(namespace)
+      Print.printWith(path, config, new IntWritable, Writables.bytesWritable(4096))(print(createMutableFact, converter, mapper) _)
```
charleso: Did we decide whether `IntWritable` or `BytesWritable` would be better?

ivory-storage/src/main/scala/com/ambiata/ivory/storage/fact/MrFactConverter.scala
```diff
@@ -0,0 +1,44 @@
...
+  def convert(fact: MutableFact, key: NullWritable, value: BytesWritable): Unit = {
+    deserialiser.fromBytesViewUnsafe(fact, value.getBytes, 0, value.getLength)
+    ()
+  }
+}
+
+/** Partitioned sequence files where the key is a date and the value is thrift serialised ThriftFact bytes.
+    The partition is the namespace. */
+case class NamespaceDateFactConverter(namespace: Namespace) extends MrFactConverter[IntWritable, BytesWritable] {
+  val deserialiser: ThriftSerialiser = ThriftSerialiser()
```
charleso: I would probably loose the type here - seems redundant? /supernit

charleso:
:wind_chime:

raronson:
@markhibberd @charleso I have finished making changes based off the comments, can you please take a look when you get a chance? I have also managed to get rid of all the duplication around setting up fact converters for the different format versions (thanks for the trait idea @charleso).

ivory-storage/src/main/scala/com/ambiata/ivory/storage/lookup/ReducerLookups.scala
markhibberd: Is this the _same_ behaviour with different code, or different behaviour?

ivory-storage/src/test/scala/com/ambiata/ivory/storage/legacy/SnapshotLoader.scala
```diff
@@ -0,0 +1,25 @@
...
+import com.ambiata.ivory.core._
+import com.ambiata.ivory.core.thrift._
+import com.ambiata.ivory.mr.FactFormats._
+
+import com.nicta.scoobi.Scoobi._
+import org.apache.hadoop.fs.Path
+
+/* TODO Remove
+ * This is a temp solution to reading v1 and v2 snapshot formats in tests. It should be replaced
+ * with notion SequenceUtil when it can handle multiple key/value types in sequence files */
```
markhibberd: I don't understand this comment, why would a sequence file have multiple different types?

markhibberd:
Two minor questions, happy to just discuss, but it looks really good. :hamburger:

ivory-storage/src/main/scala/com/ambiata/ivory/storage/lookup/ReducerLookups.scala
charleso: > Is this the same behaviour with different code, or different behaviour?

I can't see any problem with this. I can't see anywhere else that relies on the exact ordering of dictionary features not being sorted.

markhibberd:
:ship:

ivory-storage/src/main/scala/com/ambiata/ivory/storage/fact/Namespaces.scala
```diff
@@ -34,4 +36,11 @@ object Namespaces {
     sizes.foldLeft(Map[Namespace, BytesQuantity]()) {
       case (k, v) => k + (v._1 -> implicitly[Numeric[BytesQuantity]].mkNumericOps(k.getOrElse(v._1, 0.mb)).+(v._2))
    }
+
+  def fromSnapshotPath(path: Path): Namespace = {
+    Namespace.nameFromStringDisjunction(path.getParent.getName) match {
+      case scalaz.\/-(n) => n
+      case scalaz.-\/(e) => Crash.error(Crash.DataIntegrity, s"Can not parse snapshot namespace from path '${path}' - ${e}")
```
charleso: Nit, but why the redundant namespacing here? I'm a fan of `fold`. ;-)

ivory-operation/src/main/scala/com/ambiata/ivory/operation/display/PrintFacts.scala
```diff
@@ -1,52 +1,63 @@
...
-    IO.putStrLn(renderFact(delim, tombstone, f))
+  def printFactset(repository: Repository, factset: Factset, entities: List[String], attributes: List[String]): RIO[String \/ Unit] = {
+    val mapper = DumpFactsMapper(entities.toSet, attributes.toSet, s"Factset[${factset.id.render}]")
+    repository match {
+      case hr @ HdfsRepository(_) => for {
+        paths <- IvoryInputs.factsetPaths(hr, factset).traverse(expandPath).map(_.flatten).run(hr.configuration)
+        ret   <- paths.traverse(path => {
+          val converter = factset.format match {
+            case FactsetFormat.V1 => MrFactsetFactFormatV1.factConverter(path)
+            case FactsetFormat.V2 => MrFactsetFactFormatV2.factConverter(path)
```
charleso: At some point it might be handy to have a `MrFactsetFactFormat.converterFromFormat` function if we do this elsewhere.

ivory-storage/src/main/scala/com/ambiata/ivory/storage/fact/MrFactsetFactFormat.scala
```diff
@@ -0,0 +1,22 @@
...
+trait MrFactsetFactFormatV1 extends MrFactFormat[NullWritable, BytesWritable] {
+  val format: FactsetFormat = FactsetFormat.V1
+  def factConverter(path: Path): MrFactConverter[NullWritable, BytesWritable] =
+    PartitionFactConverter(FactsetInfo.getBaseInfo(path)._2)
+}
+
+object MrFactsetFactFormatV1 extends MrFactsetFactFormatV1
+
+trait MrFactsetFactFormatV2 extends MrFactFormat[NullWritable, BytesWritable] {
+  val format: FactsetFormat = FactsetFormat.V2
```
charleso: FYI This isn't declared on `MrFactFormat` but just happens to work in the MR mappers due to traitness.

ivory-storage/src/main/scala/com/ambiata/ivory/storage/fact/Namespaces.scala
raronson: Im not sure why i added the `scalaz` namespace. I should use fold more, old habits...

charleso:
:izakaya_lantern:

ivory-storage/src/main/scala/com/ambiata/ivory/storage/fact/MrFactsetFactFormat.scala
raronson: Yeah, i wasn't entirely happy about this, but im not sure what else to do, other then declaring `format` in the mapper class.

PR #576, opened 2015-01-22
Add 'tmp' level to Repository.tmp

raronson:
I don't know how i missed this, but it looks like anytime `Repository.tmp` was called it gave a path __not__ under the repository tmp dir.


charleso:
:whale:

raronson:
Looking through the code base, these are the possible top level dirs which temp data might have been created under:
  * chord
  * extraction
  * squash
  * identifier
  * recreate
  * expired

PR #581, opened 2015-01-28
Fix extracting to multiple locations/formats

charleso:
@nhibberd does this look right?

/cc  @domdere


domdere:
@markhibberd this is the problem i was encountering that was stopping me from doing an extract with multiple outputs

nhibberd:
The issue was multiple output formats running at the same time and crashing because of conflicting folders?
:car:

charleso:
@nhibberd It re-used the same temporary output folder, and so it would crash on the second output file.

PR #590, opened 2015-02-03
Add union reducer

FrontierOtr:
So unfortunately good specs for this one were not really possible for a few reasons.
1) CLI test can't work as list value imports aren't supported in text.
2) Good expression and dictionary specs can't work because filtering on lists isn't supported, so DefinitionWithQuery (which has a get on a option?) doesn't work either.


markhibberd:
We should add support for json ingest / extract so we can add the tests before we merge.

ivory-operation/src/test/scala/com/ambiata/ivory/operation/extraction/reduction/UnionReducerSpec.scala
```diff
@@ -0,0 +1,26 @@
...
+      xs.flatten.toSet.size
+  )
+
+  def unions = prop((xs: List[List[Int]]) =>
+    ReducerUtil.run(new UnionReducer[Int], xs).toSet ====
+      xs.flatten.toSet
+  )
+
+  // Overriding the List arbitrary so the test completes in a sane amount of time.
+  implicit def LittleList: Arbitrary[List[Int]] = Arbitrary(for {
```
charleso: Yeah don't do that.

You can specify the ScalaCheck "size" on the prop, which means that lists/strings will be capped. Preferably you can just use [List10](https://github.com/ambiata/disorder/blob/master/src/main/scala/com/ambiata/disorder/List10.scala) or [List100](https://github.com/ambiata/disorder/blob/master/src/main/scala/com/ambiata/disorder/List100.scala).

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/reduction/UnionReducer.scala
```diff
@@ -0,0 +1,16 @@
+package com.ambiata.ivory.operation.extraction.reduction
+
+/** Handle the latest of a single struct value */
+class UnionReducer[A] extends ReductionFold[Set[A], List[A], List[A]] {
+
+  def initial: Set[A] = Set[A]()
```
charleso: In general, I don't know what the performance/GC is going to be like for this reducer, with all the conversions to scala `List` and using `Set` (and then back to `List` and then to `j.u.List`)

ivory-core/src/main/scala/com/ambiata/ivory/core/Expression.scala
```diff
@@ -134,6 +136,10 @@ object Expression {
     }
     (exp match {
       case Count                         => ok
+      case Union                         => encoding match {
+        case ListEncoding(StringEncoding | IntEncoding | DateEncoding) => ok
+        case _ => "Unsupported format for Union expression".left
```
charleso: We should list all the other encodings here

charleso:
Yeah, not having a full end-to-end cli test for this and `latest_n` is a little worrying. The other option is to write a hard-coded Scala test, although having the JSON input/output will be generally handy in lots of ways.

markhibberd:
We need JSON in/out asap. I don't any more reducers merged before we have it, so :-1: to work around.

markhibberd:
@FrontierOtr Go for it.

charleso:
Do we need to bump the metadata version as part of this change?

markhibberd:
@charleso I don't think so all the expressions are text in the dictionary right?

charleso:
@FrontierOtr You're going to have to bump the version again (or the cli tests will fail)

ivory-operation/src/main/scala/com/ambiata/ivory/operation/update/Update.scala
```diff
@@ -19,14 +19,16 @@ object Update {
...
       case MetadataVersion.V2 =>
-        UpdateV2.update >> incrementVersion(config, MetadataVersion.V4)
+        UpdateV2.update >> incrementVersion(config, MetadataVersion.V5)
       case MetadataVersion.V3 =>
-        incrementVersion(config, MetadataVersion.V4)
+        incrementVersion(config, MetadataVersion.V5)
       case MetadataVersion.V4 =>
+        incrementVersion(config, MetadataVersion.V5)
+      case MetadataVersion.V5 =>
         RepositoryT.fromRIO(_ => RIO.putStrLn("Repository is already at latest version [v4]"))
```
charleso: s/v4/v5/

charleso:
Add an entry to `versions.sh`

charleso:
:+1:

PR #591, opened 2015-02-04
Add json ingest from structs and lists, and change extract to use output...

FrontierOtr:
... json structs and lists


ivory-storage/src/main/scala/com/ambiata/ivory/storage/parse/EavtParsers.scala
```diff
@@ -48,7 +48,7 @@ object EavtParsers {
   def valueFromString(meta: ConcreteDefinition, raw: String): Validation[String, Value] = meta.encoding match {
```
markhibberd: In this file, I don't think it is calling split on delimiter, I don't think this is safe any more, it should count forward delims for entity and attribute, and backwards for time, and treat the middle as the value (due to potential | in json.

ivory-core/src/main/scala/com/ambiata/ivory/core/Fact.scala
```diff
@@ -217,6 +229,46 @@ object Value {
...
+        ls <- sub match {
+          case se: StructEncoding    => jl.traverseU(v => parseJsonStruct(se, v))
+          case pe: PrimitiveEncoding => jl.traverseU(v => parseJsonPrimitive(pe, v))
+        }
+      } yield ListValue(ls)
+    }
+    (for {
+      json  <- raw.decodeOption[Json].toRightDisjunction(s"Value '$raw' is not a List")
+      value <- encoding match {
+        case pe: PrimitiveEncoding => parseJsonPrimitive(pe, json)
```
markhibberd: *shakes fist at @charleso* I don't think I can ever forgive you these interfaces.

ivory-storage/src/main/scala/com/ambiata/ivory/storage/parse/EavtParsers.scala
```diff
@@ -48,7 +48,7 @@ object EavtParsers {
...
     case _ if meta.tombstoneValue.contains(raw)  => TombstoneValue.success[String]
     case p: PrimitiveEncoding                    => Value.parsePrimitive(p, raw)
-    case s: StructEncoding                       => "Struct encoding not supported".failure
-    case _: ListEncoding                         => "List encoding not supported".failure
+    case s: StructEncoding                       => Value.parseJson(s, raw)
+    case l: ListEncoding                         => Value.parseJson(l, raw)
```
markhibberd: Is there an arbitrary & spec to update for these as well?

markhibberd:
Need to add something to release notes about breaking output format. I need to think a bit more about if it is ok to just change the output format like that, it will affect some of our downstream code, but I will need to talk to @blever if it will hit anyone else (I don't think so though).

ivory-core/src/main/scala/com/ambiata/ivory/core/Fact.scala
```diff
@@ -217,6 +229,46 @@ object Value {
     case StringEncoding                          => StringValue(raw).success[String]
   }
 
+  def parseJson(encoding: Encoding, raw: String): Validation[String, Value] = {
```
charleso: As mentioned IRL I might be tempted to extract this to it's own file/class.

ivory-storage/src/main/scala/com/ambiata/ivory/storage/parse/EavtParsers.scala
charleso: > Is there an arbitrary & spec to update for these as well?

+1

charleso:
@markhibberd I've pushed up what I have for feedback (and builds). Just digging in to the keyed_set problem now.

FrontierOtr:
Did you squash my commits out? 
:(

PR #592, opened 2015-02-04
Fix broken invalid-dictionary cli test

raronson:


markhibberd:
If the test passes now :sheep: it

PR #594, opened 2015-02-05
Property express Encoding types without inheritance

charleso:
Complain and ye shall receive.

The names of the new types are kinda sucky. Happy to rename if people have suggestions.

Not sure about the last two commits and the change to `StructEncodedValue` - but that made it possible to add `foldRec`.

Depending on how this is received I'm happy to do `Value` next.

### DIFF
--- a/ivory-core/src/main/scala/com/ambiata/ivory/core/Encoding.scala
+++ b/ivory-core/src/main/scala/com/ambiata/ivory/core/Encoding.scala
@@ -7,6 +7,16 @@ sealed trait Encoding {
     case EncodingStruct(e) => s(e)
     case EncodingList(e) => l(e)
   }
+
+  def foldRec[A](p: PrimitiveEncoding => A, s: Map[String, StructEncodedValue[A]] => A, l: A => A): A =
+    fold(
+      p,
+      e => s(e.values.mapValues(_.map(p))),
+      el => l(el.encoding.fold(
+        p,
+        e => s(e.values.mapValues(_.map(p)))
+      ))
+    )
 }
 case class EncodingPrim(encoding: PrimitiveEncoding) extends Encoding
 case class EncodingStruct(encoding: StructEncoding) extends Encoding
@@ -68,17 +78,17 @@ object StructEncodedValue {
 
 object Encoding {
 
-  def render(enc: Encoding): String = enc match {
-    case EncodingList(ListEncoding(e)) => "[" + renderSub(e) + "]"
-    case EncodingSub(e) => renderSub(e)
-  }
-
-  private def renderSub(enc: SubEncoding): String = enc match {
-    case SubPrim(e) => renderPrimitive(e)
-    case SubStruct(m)    => "(" + m.values.map {
-      case (n, v) => n + ":" + renderPrimitive(v.encoding) + (if (v.optional) "*" else "")
-    }.mkString(",") + ")"
-  }
+  def render(enc: Encoding): String = enc.foldRec[String](
+    renderPrimitive,
+    m =>
+      "(" +
+        m.map {
+          case (n, v) => n + ":" + v.encoding + (if (v.optional) "*" else "")
+        }.mkString(",") +
+      ")",
+    e =>
+      "[" + e + "]"
+  )
 
   def renderPrimitive(enc: PrimitiveEncoding): String = enc match {
     case BooleanEncoding => "boolean"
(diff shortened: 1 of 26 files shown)


ivory-core/src/main/scala/com/ambiata/ivory/core/Encoding.scala
```diff
@@ -12,39 +34,53 @@ case object DoubleEncoding    extends PrimitiveEncoding
...
+case class ListEncoding(encoding: SubEncoding) {
+
+  def toEncoding: Encoding =
+    EncodingList(this)
+}
 
 // NOTE: For now we don't support nested structs
-case class StructEncodedValue(encoding: PrimitiveEncoding, optional: Boolean = false) {
-  def opt: StructEncodedValue =
+case class StructEncodedValue[A](encoding: A, optional: Boolean = false) {
```
raronson: Why is this `A` and not constrained to `Encoding` or `PrimitiveEncoding`? /curious

charleso: > Why is this A and not constrained to Encoding or PrimitiveEncoding? /curious

Just because of the `foldRec` above. If you want to have the primtive encodings returned as something else, you then need to represent "that returned value plus isOptional". If/when we add more values to the struct they would want to be included in that value. Which is basically `StructEncodedValue` without the constraint. As mentioned though, I'm happy not to do this part, but having `foldRec` seemed handy.

ivory-core/src/main/scala/com/ambiata/ivory/core/Encoding.scala
```diff
@@ -1,10 +1,32 @@
...
+sealed trait Encoding {
 
-sealed trait Encoding
+  def fold[A](p: PrimitiveEncoding => A, s: StructEncoding => A, l: ListEncoding => A): A = this match {
+    case EncodingList(e) => l(e)
+    case EncodingSub(SubStruct(e)) => s(e)
+    case EncodingSub(SubPrim(e)) => p(e)
+  }
+
+  def foldRec[A](p: PrimitiveEncoding => A)(s: Map[String, StructEncodedValue[A]] => A, l: A => A): A = {
```
markhibberd: can this just be a normal param list, any reason for splitting it?

charleso: > can this just be a normal param list, any reason for splitting it?

Only for type-inference. :( Happy to change it if you prefer, but it just makes consuming it easier.

ivory-core/src/main/scala/com/ambiata/ivory/core/Encoding.scala
markhibberd: Snip the default param while you are there.

charleso: > Snip the default param while you are there.

With pleasure.

markhibberd:
Guilt is a wonderful thing.

Structure of it all looks ok, but I would like to see a concise diagram of the tree of types that is trying to be expressed. This will (a) make it easier to review, (b) document the structure a bit better, (c) gives us a place to document current limitations and potential changes (d) give us a single spot to bikeshed names (e) all of the above.

ivory-core/src/main/scala/com/ambiata/ivory/core/Encoding.scala
markhibberd: It irks me, possibly more than it should, but changing calling conventions based on particular consumes never ends as well as is intended. I vote, make a call on how it should be, and make it uniform. IMO fold and foldRec should have the same calling convention and it should either be, `(a, b, c)`, or `(a)(b)(c)`, this just feels like the middle of know where as it is.

charleso:
> Structure of it all looks ok, but I would like to see a concise diagram of the tree of types that is trying to be expressed.

I've added a very primitive one in the `Encoding` scaladoc - happy to expand although not sure how best to express it. Let me know if that all looks good. I might tackle `Value` tomorrow (or some time next week).

ivory-core/src/main/scala/com/ambiata/ivory/core/Encoding.scala
charleso: > It irks me, possibly more than it should, but changing calling conventions based on particular consumers never ends as well as is intended. 

Done. I've gone with `(a, b, c)` and consumers of `foldRec` will have to specify the generic type.

charleso:
@markhibberd Look better?

markhibberd:
:gun:

domdere:
:angel: :boom: :gun:

PR #595, opened 2015-02-05
hack

markhibberd:
ignoring the dodgy ascii art, i think I would expect something closer to this. I think the conflation of the top level sibling and nested view of  Struct/Primitive makes this far more unwieldly than it needs to be. 

I guess this really coming from a different set of first principles - how I arrived at the above was modelling the problem directly as data types - this assumes that data types are cheap, I am not really sure I understand the desire to re-use/hijack things from different contexts.
.


markhibberd:
@charleso Grab me to discuss this when you have 5.

PR #596, opened 2015-02-09
Externalise FeatureIds - 1. Store FeatureId mappings

raronson:


ivory-core/src/main/scala/com/ambiata/ivory/core/FeatureIdMappings.scala
```diff
@@ -0,0 +1,30 @@
+package com.ambiata.ivory.core
+
+case class FeatureIdIndex(int: Int) extends AnyVal
+
+case class FeatureIdMappings(private val lookup: Array[FeatureId]) {
+
+  def get(index: FeatureIdIndex): Option[FeatureId] =
+    if(index.int >= lookup.length) None else Some(lookup(index.int))
```
charleso: `(index.int < lookup.length).option(lookup(index.int))` /syntax

ivory-core/src/test/scala/com/ambiata/ivory/core/IvoryLocationSpec.scala
```diff
@@ -100,6 +101,14 @@ IvoryLocation
     } must beOkValue(contents)
   }
 
+  def readWriteBytes = { temporaryType: TemporaryType =>
+    val bytes: Array[Byte] = "a\nb \u9999\n".getBytes
```
markhibberd: Prop?

ivory-core/src/test/scala/com/ambiata/ivory/core/gen/GenIdentifier.scala
```diff
@@ -13,6 +13,9 @@ object GenIdentifier {
     name <- GenString.sensible
   } yield FeatureId(ns, name)
 
+  def featureMappings: Gen[FeatureIdMappings] =
+    GenPlus.listOfSized(1, 100, feature).map(FeatureIdMappings.apply)
```
markhibberd: Do you need to make sure these are unique?

ivory-operation/src/test/scala/com/ambiata/ivory/operation/ingestion/IngestSpec.scala
```diff
@@ -115,6 +116,23 @@ class IngestSpec extends Specification with ScalaCheck { def is = section("mr")
     } must beOkValue(facts.facts.map(_.toThrift).toSet -> (badFacts.size + 1))
   }.set(minTestsOk = 3, maxDiscardRatio = 10)
 
+  def featureIdMappings = prop((facts: PrimitiveSparseEntities, tt: TemporaryType) => {
+    val expected: List[FeatureId] = FeatureIdMappings.fromDictionary(facts.dictionary).featureIds
+    withRepository(Hdfs) { repository: Repository =>
```
charleso: These don't need to be nested any more I'm pretty sure?

ivory-storage/src/test/scala/com/ambiata/ivory/storage/metadata/FeatureIdMappingsStorageSpec.scala
```diff
@@ -0,0 +1,40 @@
...
+-------------------------------
+
+  Symmetric read / write with Key                              $symmetricKey
+  Symmetric read / write with IvoryLocation                    $symmetricIvoryLocation
+
+"""
+  val key: Key = Repository.root / FeatureIdMappingsStorage.keyname
+
+  def symmetricKey = propNoShrink((mappings: FeatureIdMappings) =>
+    RepositoryBuilder.using(repository => {
```
charleso: Again, you can use `RepositoryBuilder.repository` now to flatten this /syntax

ivory-storage/src/main/scala/com/ambiata/ivory/storage/metadata/FeatureIdMappingStorage.scala
```diff
@@ -0,0 +1,50 @@
...
+    bytes <- featureIdMappingsToBytes(mappings)
+    _     <- IvoryLocation.writeBytes(location, bytes)
+  } yield ()
+
+  def bytesToFeatureIdMappings(bytes: Array[Byte]): RIO[FeatureIdMappings] = for {
+    mappings <- RIO.safe(ThriftSerialiser().fromBytes1(() => new ThriftFeatureIdMappings, bytes))
+    features <- RIO.fromDisjunctionString(mappings.getFeatures.asScala.toList.traverseU(FeatureId.parse))
+  } yield FeatureIdMappings(features)
+
+  def featureIdMappingsToBytes(mappings: FeatureIdMappings): RIO[Array[Byte]] = {
```
charleso: Is this really `RIO`?

ivory-storage/src/main/scala/com/ambiata/ivory/storage/metadata/FeatureIdMappingStorage.scala
```diff
@@ -0,0 +1,50 @@
...
+    bytes    <- IvoryLocation.readBytes(location)
+    mappings <- bytesToFeatureIdMappings(bytes)
+  } yield mappings
+
+  def toIvoryLocation(location: IvoryLocation, mappings: FeatureIdMappings): RIO[Unit] = for {
+    bytes <- featureIdMappingsToBytes(mappings)
+    _     <- IvoryLocation.writeBytes(location, bytes)
+  } yield ()
+
+  def bytesToFeatureIdMappings(bytes: Array[Byte]): RIO[FeatureIdMappings] = for {
```
charleso: Same here - this should probably return `String \/ FeatureIdMappings` or thereabouts?

ivory-core/src/test/scala/com/ambiata/ivory/core/IvoryLocationSpec.scala
raronson: I didn't add a property because there aren't any in this test and above there is this comment `// Sigh, this can be fixed with rejigging the dependencies, but IvoryLocation should go very soon`

ivory-core/src/test/scala/com/ambiata/ivory/core/gen/GenIdentifier.scala
raronson: Yep, will do

ivory-storage/src/test/scala/com/ambiata/ivory/storage/metadata/FeatureIdMappingsStorageSpec.scala
raronson: thanks, thats nicer

ivory-storage/src/main/scala/com/ambiata/ivory/storage/metadata/FeatureIdMappingStorage.scala
raronson: Yeah, good point

ivory-core/src/test/scala/com/ambiata/ivory/core/IvoryLocationSpec.scala
markhibberd: One or the other, no code, or properties ;)

I don't know what the comment is getting at though. Might make more work to do it, but worth it, especially for a round-trip test with bytes.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/ingestion/FactImporter.scala
```diff
@@ -28,6 +29,8 @@ object FactImporter {
     IvoryT.read[RIO] >>= (read => IvoryT.fromRIO { for {
       hr            <- repository.asHdfsRepository
       dictionary    <- Metadata.latestDictionaryFromIvory(repository)
+      mappings      = FeatureIdMappings.fromDictionary(dictionary)
+      _             <- FeatureIdMappingsStorage.toKeyStore(repository, Repository.factset(factsetId) / FeatureIdMappingsStorage.keyname, mappings)
```
markhibberd: Is this sync'd properly with the input dataset, it could be, I just can't see a good test for it.

ivory-core/src/main/scala/com/ambiata/ivory/core/FeatureIdMappings.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  def getUnsafe(index: FeatureIdIndex): FeatureId =
```
charleso: I'm assuming we want this for the MR jobs? Be good to have a property for it.

ivory-storage/src/main/scala/com/ambiata/ivory/storage/metadata/FeatureIdMappingStorage.scala
markhibberd: :100: should be pure.

markhibberd:
:gb: After the few minor comments.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/ingestion/FactImporter.scala
raronson: Im not sure i understand

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/squash/SquashJob.scala
```diff
@@ -52,11 +52,13 @@ object SquashJob {
...
              job: (Job, MrContext, Int), cluster: Cluster): RIO[ShadowOutputDataset] = for {
-    key    <- Repository.tmpDir("squash")
-    hr     <- repository.asHdfsRepository
-    shadow = ShadowOutputDataset.fromIvoryLocation(hr.toIvoryLocation(key))
-    prof   <- run(job._1, job._2, job._3, dictionary, shadow.hdfsPath, cluster.codec, conf, latest = true)
-    _      <- IvoryLocation.writeUtf8Lines(hr.toIvoryLocation(key) </> FileName.unsafe(".profile"), SquashStats.asPsvLines(prof))
+    key      <- Repository.tmpDir("squash")
+    hr       <- repository.asHdfsRepository
+    mappings = FeatureIdMappings.fromDictionary(dictionary)
+    _        <- FeatureIdMappingsStorage.toKeyStore(repository, key / FeatureIdMappingsStorage.keyname, mappings)
```
charleso: Maybe a silly/bad idea but do we want to make it easier/safer to save whenever we convert from a dictionary?

```
def fromDictionaryAndSave(r: Repository, key: Key,  d: Dictionary): RIO[FeatureIdMappings]
```

ivory-storage/src/main/scala/com/ambiata/ivory/storage/metadata/FeatureIdMappingStorage.scala
charleso: It's probably worth adding it's own symmetric property for these two while you're at it. :)

ivory-operation/src/main/scala/com/ambiata/ivory/operation/ingestion/FactImporter.scala
markhibberd: Ah, maybe I don't. I was possibly thinking ahead to how it works when the s3 sync happens on both source and destination because of the sync code below, but you are right my comment is irrelevant because it only needs to sync source data not the factset.

charleso:
:hand:

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/squash/SquashJob.scala
raronson: Yeah, this reduces duplicate code, will change

ivory-core/src/test/scala/com/ambiata/ivory/core/gen/GenIdentifier.scala
```diff
@@ -13,6 +13,9 @@ object GenIdentifier {
     name <- GenString.sensible
   } yield FeatureId(ns, name)
 
+  def featureMappings: Gen[FeatureIdMappings] =
+    GenPlus.listOfSized(1, 100, feature).map(_.distinct).map(FeatureIdMappings.apply)
```
charleso: Just as food for thought, I might be tempted to zip with an index and do something with that to guarantee uniqueness.

ivory-storage/src/main/scala/com/ambiata/ivory/storage/metadata/FeatureIdMappingStorage.scala
```diff
@@ -0,0 +1,57 @@
...
+
+  def featureIdMappingsToBytes(mappings: FeatureIdMappings): String \/ Array[Byte] = {
+    val serialiser = ThriftSerialiser()
+    \/.fromTryCatchNonFatal(serialiser.toBytes(new ThriftFeatureIdMappings(mappings.featureIds.map(_.toString).asJava)))
+      .leftMap(t => s"Could not serialise mappings '${mappings}' - ${t}")
+  }
+
+  def fromDictionaryAndSave(repository: Repository, base: Key, dictionary: Dictionary): RIO[FeatureIdMappings] = {
+     val mappings = FeatureIdMappings.fromDictionary(dictionary)
+     FeatureIdMappingsStorage.toKeyStore(repository, base / keyname, mappings).map(_ => mappings)
```
charleso: `.map(_ => X)` == `.as(X)` /tip

ivory-storage/src/test/scala/com/ambiata/ivory/storage/metadata/FeatureIdMappingsStorageSpec.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  def symmetricBytes = prop((mappings: FeatureIdMappings) =>
+    (for {
+      bytes     <- FeatureIdMappingsStorage.featureIdMappingsToBytes(mappings)
+      mappings2 <- FeatureIdMappingsStorage.bytesToFeatureIdMappings(bytes)
+    } yield mappings2.featureIds) ==== mappings.featureIds.right
+  )
+
+  def persistFromDictionary = prop((dictionary: Dictionary) => {
+    val base = Repository.root / "mapping-test"
```
charleso: We should try to avoid hard coding strings in properties, as convenient as it is (and I"m guilty of this as well). :)

ivory-storage/src/main/scala/com/ambiata/ivory/storage/metadata/FeatureIdMappingStorage.scala
```diff
@@ -0,0 +1,57 @@
...
+    _     <- IvoryLocation.writeBytes(location, bytes)
+  } yield ()
+
+  def bytesToFeatureIdMappings(bytes: Array[Byte]): String \/ FeatureIdMappings = for {
+    mappings <- \/.fromTryCatchNonFatal(ThriftSerialiser().fromBytes1(() => new ThriftFeatureIdMappings, bytes))
+                  .leftMap(t => s"Could not deserialise bytes - ${t}")
+    features <- mappings.getFeatures.asScala.toList.traverseU(FeatureId.parse)
+  } yield FeatureIdMappings(features)
+
+  def featureIdMappingsToBytes(mappings: FeatureIdMappings): String \/ Array[Byte] = {
```
charleso: Hmm, not sure about this type - should it just return `Array[Byte]`?

raronson: `TSerializerCopy.serialize` throws a `TException`, not sure if we need to worry about that? If not i can change.

charleso: > TSerializerCopy.serialize throws a TException, not sure if we need to worry about that?

I think it's pretty much impossible in this case (and I don't think we check anywhere else). Happy to be told it's worth checking though.

ivory-storage/src/test/scala/com/ambiata/ivory/storage/metadata/FeatureIdMappingsStorageSpec.scala
raronson: I agree, but how do you get an arbitrary `Key`?

PR #597, opened 2015-02-09
Add user documentation for ingest/extract formats

charleso:
@blever Sorry I never created the PR (and then forgot). I wasn't very happy with it at the time. Maybe if you have 5 minutes today to talk about it I'm happy to fix up and merge.


doc/user/formats.md
```diff
@@ -0,0 +1,60 @@
...
+Extract
+-------
+
+These are the formats that can be specified for both `snapshot`/`chord`.
+
+Format                    | Type   | Example                                           | Description
+------------------------- | ------ | ------------------------------------------------- | ------------
+`dense:[ESCAPING:]DELIM`  | text   | `entity|feature1|NA|feature3`                     | All features/values for a single fact are represented in a single line, in which the order is defined in an accompanying `.dictionary`.
+`sparse:[ESCAPING:]DELIM` | text   | `entity|ns|attribute|value|YYYY-MM-DD'T'HH:MM:SS` | Each fact is represented by a separate entry
+`dense:thrift`            | thrift |                                                   | Similar in concept to `dense` text, all values for a single entity are contained in a `List` of values in the same order as the included dictionary.
```
blever: Can we have a pointer to the thrift schema file for this?

doc/user/formats.md
```diff
@@ -0,0 +1,60 @@
...
+`sparse:thrift`           | thrift |                                                   | Similar in concept to `sparse` text, all values for a single entity are contains in a `Map`, where missing values are not included.
```
blever: Same here - a pointer the thrift schema file.

markhibberd: It depends on release, so can it be pointed to in the tarball.

etorreborre:
@charleso it'd be nice as well to specify in the README what is currently in the code (and give a concrete example of the expected paths): 
```scala
If NAMESPACE is set then the input path is expected to contain partitioned factsets
```

blever:
:cherries: if we can point to which thrift schema in the tarball to use for extractions

nhibberd:
@charleso bump

PR #600, opened 2015-02-10
Added doc for days_since

charleso:


doc/design/feature-gen.md
```diff
@@ -36,6 +36,12 @@ be summed.
 ### `count_unique`
 Return the number of unique fact values in `facts`.
 
+### `days_since`
+Return the number of days _after_ a date _value_ and the `ref`.
```
FrontierOtr: After doesn't make sense here, you're implying they're the same day or it's after both of them. It returns the number of days between a date value and the reference.

charleso:
@FrontierOtr Good point. Better?

doc/design/feature-gen.md
```diff
@@ -36,6 +36,12 @@ be summed.
 ### `count_unique`
 Return the number of unique fact values in `facts`.
 
+### `days_since`
+Return the number of days between a date _value_ and the `ref`.
+If the value is _after_ the `ref` then the value will be negative.
```
FrontierOtr: The result will be negative.

nhibberd:
@charleso @HuwCampbell bump

PR #601, opened 2015-02-16
Notes on requirements for BigQuery support

blever:
@markhibberd FYI - as discussed IRL

### DIFF
--- a/doc/design/big-query.md
+++ b/doc/design/big-query.md
@@ -1,0 +1,54 @@
+Big Query Support
+=================
+
+This describes additions that can be made to Ivory to make it simpler to load
+snapshots/chords in to a BigQuery table.
+
+Loading data into BigQuery
+--------------------------
+
+There are number of different ways to [preapre data for loading](https://cloud.google.com/bigquery/preparing-data-for-bigquery)
+in to BigQuery but the most appropriate for snapshots/chords will be:
+
+* 1 table per snapshot/chord
+* table schema defined in JSON - derrived from snapshot/chord dictionary file
+* snapshot/chord represented as dense JSON text files
+* Gzip compressed snapshot files
+
+
+Immediate Ivory requirements
+----------------------------
+
+### Dense JSON text output format for snapshots/chords
+
+An output format that is *row oriented* (i.e. row per entity) where each row is a JSON object.
+For example:
+
+```
+uid  |  a: int  | b: double  | c: string  | d: [string]     | e: {name: string, age: int}
+-----|----------|------------|------------|-----------------|-----------------------------
+u111 |     3    |   0.67     |   "foo"    |       -         |               -            
+u222 |     4    |    -       |   "bar"    |       -         |               -            
+u333 |     5    |    -       |     -      | ["foo","bar"]   |               -            
+u444 |     6    |    -       |     -      |       -         | {"name": "bob", "age": 42 }
+```
+
+Would be represented as:
+
+```
+{"uid": "u111", "a": 3, "b": 0.67, "c": "foo"}
+{"uid": "u222", "a": 4, "c": "bar"}
+{"uid": "u333", "a": 5, "d": ["foo", "bar"]}
+{"uid": "u444", "a": 6, "e": {"name": "bob", "age": 42}}
+```
+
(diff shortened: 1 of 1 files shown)


doc/design/big-query.md
```diff
@@ -0,0 +1,54 @@
...
+{"uid": "u111", "a": 3, "b": 0.67, "c": "foo"}
+{"uid": "u222", "a": 4, "c": "bar"}
+{"uid": "u333", "a": 5, "d": ["foo", "bar"]}
+{"uid": "u444", "a": 6, "e": {"name": "bob", "age": 42}}
+```
+
+Notes:
+  * string values containing newlines will need to be escaped so that the output is a JSON
+  record per line
+  * for feature values that are *dates*, the output will need to comply the data/time formatting
```
markhibberd: These are just dates, not date/times, which makes this a bit questionable. Might need a separate discussion for this, but I am not sure it is a good idea to make up a 00:00 time component or similar by default, might want to specify what to do in the output format instead of "random stuff happens".

doc/design/big-query.md
```diff
@@ -0,0 +1,54 @@
...
+  supported by BigQuery - https://cloud.google.com/bigquery/preparing-data-for-bigquery#datatypes
+  * feature names should include their namespace as well, e.g. `"namespace:name": "...`
```
markhibberd: Is this an absolute requirement from an ease of querying requirement? Otherwise pulling out the namespace for ease of navigation and reduction of string-duplication might be an idea.

markhibberd:
:sake:

doc/design/big-query.md
blever: It's not an absolute requirement. If we can ensure that feature names are unique across all namespaces then it's probably ok to drop. Is that the case?

markhibberd: It isn't guaranteed if that is what you mean. I was more referring to something like:
```
{"entity": ...., "namespace1": { "featureX": ..., "featureY": ...}, "namespace2": { "featureZ" ... }
```

blever: I don't think that will work for BQ.

Any reason why we can't introduce the requirement of unique feature names?

markhibberd: Ok. I am ok with unique feature names.

doc/design/big-query.md
```diff
@@ -0,0 +1,56 @@
...
+{"uid": "u444", "a": 6, "e": {"name": "bob", "age": 42}}
+```
+
+Notes:
+  * string values containing newlines will need to be escaped so that the output is a JSON
+  record per line
+  * for feature values that are *dates*, the output will need to comply the data/time formatting
+  supported by BigQuery - https://cloud.google.com/bigquery/preparing-data-for-bigquery#datatypes
+  * feature names will not include their namespace, just name, as Ivory will ensure these are
+  globally unique
```
blever: Dropping the requirement to output namespace.

blever:
@markhibberd incorporated comments

markhibberd:
Looks good, we will just need to add unique constraint and evaluate effect on downstream repos. But either way will be able to make it work. Merge away. :fast_forward:

blever:
Capture the globally unique attribute names in an issue - https://github.com/ambiata/ivory/issues/602.

doc/design/big-query.md
```diff
@@ -0,0 +1,58 @@
+Big Query Support
+=================
+
+This describes additions that can be made to Ivory to make it simpler to load
+snapshots/chords in to a BigQuery table.
+
+Loading data into BigQuery
+--------------------------
+
+There are number of different ways to [preapre data for loading](https://cloud.google.com/bigquery/preparing-data-for-bigquery)
```
etorreborre: s/preapre/prepare

doc/design/big-query.md
```diff
@@ -0,0 +1,58 @@
...
+in to BigQuery but the most appropriate for snapshots/chords will be:
+
+* 1 table per snapshot/chord
+* table schema defined in JSON - derrived from snapshot/chord dictionary file
```
etorreborre: s/derrived/derived

PR #603, opened 2015-02-23
LatestN wasn't idempotent

charleso:
I'm happy to go through all the specs and add this for all the `Reduction`s

### DIFF
--- a/ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/reduction/LatestNReducer.scala
+++ b/ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/reduction/LatestNReducer.scala
@@ -59,7 +59,11 @@ class LatestNStructReducer[@specialized(Int, Float, Double, Boolean) A : scala.r
 
   val start = ClockState[A](n, Array.fill(n)(seed), 0, 0)
 
-  def initial: ClockState[A] = start
+  def initial: ClockState[A] = {
+    start.hand  = 0
+    start.count = 0
+    start
+  }
 
   def fold(a: ClockState[A], value: A): ClockState[A] = {
     a.tick(value)
--- a/ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/reduction/LatestNReducer.scala
+++ b/ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/reduction/LatestNReducer.scala
@@ -22,16 +22,19 @@ case class ClockState[@specialized(Int, Float, Double, Boolean) A](val n: Int, v
     }
     xs.reverse
   }
+
+  def reset(): Unit = {
+    hand = 0
+    count = 0
+  }
 }
 
 class LatestNReducer(n: Int) extends Reduction {
 
   val a = ClockState[ThriftFactValue](n, new Array[ThriftFactValue](n), 0, 0)
 
-  def clear(): Unit = {
-    a.hand  = 0
-    a.count = 0
-  }
+  def clear(): Unit =
+    a.reset()
 
   def update(fv: Fact): Unit = {
     val tv = fv.toThrift.getValue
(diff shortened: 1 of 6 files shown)


ivory-operation/src/test/scala/com/ambiata/ivory/operation/extraction/reduction/ReducerUtil.scala
```diff
@@ -32,6 +32,9 @@ object ReducerUtil {
     r.aggregate(dates)
   }
 
+  def idempotent[A, B, C](r: ReductionFold[A, B, C], l: List[B]): Boolean =
+    run(r, l) == run(r, l)
```
charleso: Wondering if I should return a `Matcher` from this?

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/reduction/LatestNReducer.scala
```diff
@@ -59,7 +59,11 @@ class LatestNStructReducer[@specialized(Int, Float, Double, Boolean) A : scala.r
 
   val start = ClockState[A](n, Array.fill(n)(seed), 0, 0)
 
-  def initial: ClockState[A] = start
+  def initial: ClockState[A] = {
+    start.hand  = 0
+    start.count = 0
```
markhibberd: `reset` on ClockState perhaps, remove risk of this or clear being forgotten if this ever changes.

charleso: > reset on ClockState perhaps,

Yep, I was already thinking about it, not sure why I didn't just add it.

ivory-operation/src/test/scala/com/ambiata/ivory/operation/extraction/reduction/ReducerUtil.scala
```diff
@@ -32,6 +32,9 @@ object ReducerUtil {
     r.aggregate(dates)
   }
 
+  def idempotent[A, B, C](r: ReductionFold[A, B, C], l: List[B]): Boolean =
```
charleso: Actually that's the wrong name really. Suggestions?

markhibberd: Agree it is wrong name, no real suggestions. `repeatable`, `consistent`, `not-borked`.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/reduction/LatestNReducer.scala
```diff
@@ -22,16 +22,19 @@ case class ClockState[@specialized(Int, Float, Double, Boolean) A](val n: Int, v
     }
     xs.reverse
   }
```
markhibberd: Since the bottom of this creeped onto diff, I have a large preference for this not mutating the state, it just seem unnecessary. I kind of see why it does it, there becomes some sort of symmetrical relationship between tick and unwind, but I don't think it is that valuable and it really hurts some obvious tests, i.e. calling `save` twice should give the same answer IMO.

ivory-operation/src/test/scala/com/ambiata/ivory/operation/extraction/reduction/ReducerUtil.scala
markhibberd: I think there is a strong case for a test along the lines of:
```
 { val a = run(r, l); run(r, xxxx); val b = run(r, l;); a ==== b }
```
There should never be interference. I can see how the below covers it for the current code, but I can see how a bug could slip in that passes the simple run twice tests, but doesn't pass the interleave test.

markhibberd:
If you are happy with the suggested changes and extra test, I think we should merge, but we need to get the json PR merged and CLI tests added quickly after - get the impression they would of had a better chance of catching this.

markhibberd:
@charleso Merging this once I have a green build.

PR #609, opened 2015-02-24
Introduce reducer consecutive and interleaved laws

charleso:
"Laws" is used a little liberally here (happy to rename if we think it's an abuse of the term).

I ran out of time for `LatestByReducer`.

### DIFF
--- a/ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/reduction/KeyValue.scala
+++ b/ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/reduction/KeyValue.scala
@@ -21,4 +21,11 @@ class KeyValue[K, V]() {
     map.put(k, v)
     ()
   }
+
+  /** Just for testing */
+  override def equals(other: Any): Boolean =
+    other match {
+      case that: KeyValue[K, V] => map == that.map
+      case _ => false
+    }
 }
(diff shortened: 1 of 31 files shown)


ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/reduction/KeyValue.scala
```diff
@@ -21,4 +21,11 @@ class KeyValue[K, V]() {
     map.put(k, v)
     ()
   }
+
+  /** Just for testing */
+  override def equals(other: Any): Boolean =
```
etorreborre: It is worth overriding `hashCode` as well then.

ivory-operation/src/test/scala/com/ambiata/ivory/operation/extraction/reduction/MeanInDaysReducerSpec.scala
```diff
@@ -16,4 +18,10 @@ class MeanInDaysReducerSpec extends Specification with ScalaCheck { def is = s2"
     ReducerUtil.reduceDates(doc, new MeanInWeeksReducer) ====
       (if (doc.noOfWeeksFloor != 0) doc.datesWithFullWeeks.map(_._1).sum.toDouble / doc.noOfWeeksFloor else 0)
   })
+
+  def meanInDaysLaws =
+    ReducerUtil.reduceDatesLaws(_ => new MaximumInDaysReducer)
```
etorreborre: Shouldn't that be `new MeanInDaysReducer`?

markhibberd:
:loud_sound: after `MeanInDaysReducer` typo.

PR #614, opened 2015-02-25
Invalid snapshop if a feature mode changes

charleso:


ivory-storage/src/main/scala/com/ambiata/ivory/storage/plan/SnapshotPlan.scala
```diff
@@ -112,7 +112,8 @@ object SnapshotPlan {
    */
   def isValid(at: Date, commit: Commit, snapshot: Snapshot): Boolean =
     snapshot.date.isBeforeOrEqual(at) && snapshot.store.subsetOf(commit.store) &&
-      (sameDictionary(commit, snapshot) || noWindows(commit, snapshot) || containedBy(at, commit, snapshot))
+      (sameDictionary(commit, snapshot) || noWindows(commit, snapshot) || containedBy(at, commit, snapshot)) &&
+      sameMode(commit, snapshot)
```
markhibberd: sameDictionary should be on the outside of this I think, i.e. `sameDictionary || ((nowindows || containedby) && samemode)`

markhibberd:
:keycap_ten: After small change.

PR #615, opened 2015-02-25
Use readUtf8 for SyncExtractSpec.file

charleso:
/cc @nhibberd 

https://ci.ambiata.com/job/ivory.cdh4/365/

```
04:22:06 [error]  A counter-example is [, Posix] (after 5 tries)
04:22:06 [error]  '(true,)' is not equal to '(true,)' (RIOMatcher.scala:48)
```


nhibberd:
:car:

charleso:
For the record it's `\f` that's breaking. Just mentioning that in case this isn't handled in notion/poacher.

PR #616, opened 2015-02-25
Introduce Mode.KeyedSet

charleso:


ivory-core/src/main/thrift/core.thrift
```diff
@@ -146,6 +151,7 @@ struct ThriftDictionaryFeatureMeta {
     4: list<string> tombstoneValue;
     5: optional ThriftDictionaryFeatureValue value;
     6: optional ThriftDictionaryMode mode;
+    7: optional ThriftDictionaryModeV2 modeV2;
```
charleso: I think there's a lesson in here about avoiding thrift enums unless it's _really_ an enum.

ivory-core/src/main/scala/com/ambiata/ivory/core/Mode.scala
```diff
@@ -6,24 +6,21 @@ sealed trait Mode {
...
   }
 
   def render: String =
-    fold("state", "set")
-
-  def isState: Boolean =
-    fold(true, false)
-
-  def isSet: Boolean =
-    fold(false, true)
```
charleso: Not sure how you'll feel about this - seems like they're potentially dangerous functions if not used correctly. Thoughts?

ivory-core/src/main/scala/com/ambiata/ivory/core/Mode.scala
```diff
@@ -6,24 +6,21 @@ sealed trait Mode {
...
+    fold("state", "set", "keyed_set," + _)
 }
 
 object Mode {
   case object State extends Mode
   case object Set extends Mode
+  case class KeyedSet(key: String) extends Mode
```
charleso: Wasn't sure whether to newtype that String or not?

markhibberd: I wouldn't have a newtype for this specifically, but there is probably a more general "FieldName" or similar we may want to add for use in a few places. Not sure that I particularly mind either way at this point.

charleso: > I wouldn't have a newtype for this specifically, but there is probably a more general "FieldName" or similar we may want to add for use in a few places.

Yeah that's what I was thinking. I'm happy to leave that for now though, would prefer to get stuck in to the MR jobs tomorrow.

ivory-core/src/main/scala/com/ambiata/ivory/core/Mode.scala
markhibberd: Meh, they do what they say on the tin. If you prefer to remove them rather than add the first case I can live with that.

markhibberd:
:haircut:

PR #617, opened 2015-02-25
Validate that keyed sets have struct encoding with key

charleso:


ivory-operation/src/test/scala/com/ambiata/ivory/operation/ingestion/DictionaryImportValidateSpec.scala
```diff
@@ -21,6 +21,8 @@ class DictionaryImportValidateSpec extends Specification with ScalaCheck { def i
    is invalid with a virtual definition with an invalid filter     $virtualInvalidFilter
    is invalid with a virtual definition with an invalid expression $virtualInvalidExpression
    is invalid when there are more then Short.MaxValue features     $numFeaturesInvalid
+   is invalid when the encoding for keyed set is not a struct      $keyedSetInvalidEncoding
+   is invalid when the keyed set key is missing                    $keyedSetMissingKey
```
markhibberd: Should also be mandatory I would think, or does this cover that as well?

charleso: > Should also be mandatory I would think, or does this cover that as well?

Good point, and no this doesn't.

ivory-core/src/test/scala/com/ambiata/ivory/core/arbitraries/Arbitraries.scala
```diff
@@ -51,7 +51,7 @@ trait Arbitraries {
     Arbitrary(GenIdentifier.dictionary)
 
   implicit def EncodingArbitrary: Arbitrary[Encoding] =
-    Arbitrary(GenDictionary.encoding)
+    Arbitrary(GenDictionary.mode.flatMap(GenDictionary.encoding))
```
markhibberd: :cry: I think this should be like it was, and there should be a `.encoding` and a `.encodingWithMode` or similar on GenDictionary.

charleso: > I think this should be like it was, and there should be a .encoding and a .encodingWithMode or similar on GenDictionary.

Yeah, I did wonder about that. Happy to change.

markhibberd:
:+1: but if you feel saddened by the arbitrary cleanliness, I wouldn't be adverse to it being tightened back up.

PR #618, opened 2015-02-26
Introduce ModeReducer for accepting facts in chord/snapshot

charleso:


ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/mode/ModeReducer.scala
```diff
@@ -0,0 +1,60 @@
...
+
+import scala.collection.JavaConverters._
+
+/** Handle filtering of facts for different modes in snapshot/chord */
+trait ModeReducer {
+  /** My kingdom for decent rank-2 types */
+  type X
+  def seed: X
+  def step(state: X, fact: Fact): X
+  def accept(state: X, fact: Fact): Boolean
```
charleso: For modes it may be nicer if this was just `accept` _and_ it returned `(X, Boolean)` but I was wary of that extra allocation. Crazy town?

ivory-operation/src/test/scala/com/ambiata/ivory/operation/extraction/mode/ModeReducerSpec.scala
```diff
@@ -0,0 +1,78 @@
...
+============
+
+  Can lookup a reducer from a mode
+    $lookup
+"""
+
+  def laws(mr: ModeReducer): Prop = new Properties("ModeReducer laws") {
+    property("seed is always equal") =
+        mr.seed ?= mr.seed
+    property("accept with the same state is always equal") = forAll((f: Fact) => {
```
charleso: Yeah these aren't great - am I missing some useful ones?

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/mode/ModeReducer.scala
```diff
@@ -0,0 +1,60 @@
...
+    mode.fold(ModeReducerState, ModeReducerSet, _ => NotImplemented.keyedSet)
+
+  def fromLookup(flag: FlagLookup): Int => ModeReducer = {
+    val a = FeatureLookups.sparseMapToArray(flag.flags.asScala.map(x => x._1.intValue -> x._2.booleanValue).mapValues {
+      case true =>
+        ModeReducerSet
+      case false =>
+        ModeReducerState
+    }.toList, ModeReducerState)
+    a.apply
```
charleso: This is megamophic, should we return the `Array` instead? /crazytown

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/mode/ModeReducer.scala
markhibberd: No, I think you either need a crazy town return type or the two operations.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/mode/ModeReducer.scala
```diff
@@ -0,0 +1,60 @@
...
+  def step(state: X, fact: Fact): X
+  def accept(state: X, fact: Fact): Boolean
+}
+
+object ModeReducer {
+
+  def fromMode(mode: Mode): ModeReducer =
+    mode.fold(ModeReducerState, ModeReducerSet, _ => NotImplemented.keyedSet)
+
+  def fromLookup(flag: FlagLookup): Int => ModeReducer = {
```
markhibberd: Bit iffy on making this a function, given the context, I think the array is a fair bit nicer. My reasoning is that this isn't a nice function, it has potential for nulls and exceptions.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/mode/ModeReducer.scala
```diff
@@ -0,0 +1,60 @@
...
+  def accept(datetime: DateTime, f: Fact): Boolean =
+    datetime != f.datetime
+}
+
+object ModeReducerSet extends ModeReducer {
+
+  type X = Unit
+
+  def seed: Unit =
+    ()
```
markhibberd: What does the byte code end up looking like here? I would consider making it a val regardless.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/mode/ModeReducer.scala
```diff
@@ -0,0 +1,60 @@
...
+
+  def step(state: Unit, f: Fact): Unit =
+    ()
```
markhibberd: Consider just returning state.

ivory-operation/src/test/scala/com/ambiata/ivory/operation/extraction/mode/ModeReducerSpec.scala
```diff
@@ -0,0 +1,78 @@
...
+      val s = mr.seed
+      mr.accept(s, f) ?= mr.accept(s, f)
+    })
+    property("step with the same state is always equal") = forAll((f: Fact) => {
```
markhibberd: Not sure this one will always be true, maybe it is at the moment, but I figure the types are designed to allow crazy town, I think I would stick with something more obvious to start with:
```
mr.stap(mr.seed, f) ?= mr.stap(mr.seed, f)
```

ivory-operation/src/test/scala/com/ambiata/ivory/operation/extraction/mode/ModeReducerSpec.scala
markhibberd: Only thing would be, `seed > step > accept` is consistent.

ivory-operation/src/test/scala/com/ambiata/ivory/operation/extraction/mode/ModeReducerSpec.scala
```diff
@@ -0,0 +1,78 @@
...
+Construction
+============
+
+  Can lookup a reducer from a mode
+    $lookup
+"""
+
+  def laws(mr: ModeReducer): Prop = new Properties("ModeReducer laws") {
+    property("seed is always equal") =
+        mr.seed ?= mr.seed
```
markhibberd: Would also consider `seed > step > seed` ==== `seed` to cover off mutable state case.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/mode/ModeReducer.scala
markhibberd: Sorry, missed this comment. I think it should be Array, but not for crazy town reasons, see ^^

ivory-operation/src/test/scala/com/ambiata/ivory/operation/extraction/mode/ModeReducerSpec.scala
charleso: > Not sure this one will always be true, maybe it is at the moment, but I figure the types are designed to allow crazy town

Good point

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/mode/ModeReducer.scala
charleso: > No, I think you either need a crazy town return type or the two operations.

Fair enough. Might leave it the way it is (for now). I'm hoping/thinking this might merge in with your version eventually and we can revisit then.

charleso:
@markhibberd Thanks for the feedback. Anything else?

PR #619, opened 2015-03-02
Keyed Sets

charleso:
This is definitely not finished. There are a few `TODO`s that I've flagged which may be worth taking note of.

Pretend like I'll be passing around a `Buffer` instead of the current returning on `Array[Byte]` from `ModeKey`. I ran out of time to push that up as well and it was just too messy.

### DIFF
--- a/ivory-mr/src/main/scala/com/ambiata/ivory/mr/DictionaryCache.scala
+++ b/ivory-mr/src/main/scala/com/ambiata/ivory/mr/DictionaryCache.scala
@@ -1,0 +1,21 @@
+package com.ambiata.ivory.mr
+
+import com.ambiata.ivory.core.{Crash, Dictionary}
+import com.ambiata.ivory.core.thrift.{ThriftDictionary, DictionaryThriftConversion}
+import com.ambiata.poacher.mr.ThriftCache
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.mapreduce.Job
+
+object DictionaryCache {
+
+  val thriftCacheKey = ThriftCache.Key("dictionary")
+
+  def store(job: Job, thriftCache: ThriftCache, dictionary: Dictionary): Unit =
+    thriftCache.push(job, thriftCacheKey, DictionaryThriftConversion.dictionaryToThrift(dictionary))
+
+  def load(conf: Configuration, thriftCache: ThriftCache): Dictionary = {
+    val dictThrift = new ThriftDictionary
+    thriftCache.pop(conf, thriftCacheKey, dictThrift)
+    DictionaryThriftConversion.dictionaryFromThrift(dictThrift).fold(m => Crash.error(Crash.Serialization, m), identity)
+  }
+}
--- a/ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/ChordJob.scala
+++ b/ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/ChordJob.scala
@@ -4,7 +4,7 @@ import com.ambiata.ivory.core._
 import com.ambiata.ivory.core.thrift._
 import com.ambiata.ivory.lookup._
 import com.ambiata.ivory.operation.extraction.chord._
-import com.ambiata.ivory.operation.extraction.mode.ModeReducer
+import com.ambiata.ivory.operation.extraction.mode.{ModeKey, ModeReducer}
 import com.ambiata.ivory.operation.extraction.snapshot.SnapshotWritable
 import com.ambiata.ivory.storage.fact._
 import com.ambiata.ivory.storage.lookup._
@@ -78,8 +78,8 @@ object ChordJob {
     ctx.thriftCache.push(job, Keys.FactsetVersionLookup, FactsetLookups.versionTable(plan.datasets))
     ctx.thriftCache.push(job, Keys.FeatureIdLookup, featureIdLookup(plan.commit.dictionary.value))
     ctx.thriftCache.push(job, Keys.ChordEntitiesLookup, Entities.toChordEntities(plan.entities))
-    ctx.thriftCache.push(job, Keys.FeatureIsSetLookup, FeatureLookups.isSetTable(plan.commit.dictionary.value))
     ctx.thriftCache.push(job, Keys.ChordWindowsLookup, FeatureLookups.windowTable(plan.commit.dictionary.value))
+    DictionaryCache.store(job, ctx.thriftCache, plan.commit.dictionary.value)
 
     // run job
     if (!job.waitForCompletion(true))
@@ -106,7 +106,6 @@ object ChordJob {
     val FactsetLookup = ThriftCache.Key("factset-lookup")
     val FactsetVersionLookup = ThriftCache.Key("factset-version-lookup")
     val ChordEntitiesLookup = ThriftCache.Key("chord-entities-lookup")
-    val FeatureIsSetLookup = ThriftCache.Key("feature-is-set-lookup")
     val ChordWindowsLookup = ThriftCache.Key("chord-window-lookup")
     val Out = "out" // MultipleOutputs named output
   }
@@ -166,6 +165,7 @@ abstract class ChordFactsetMapper[K <: Writable] extends CombinableMapper[K, Byt
   val featureIdLookup = new FeatureIdLookup
 
   var entities: Entities = null
+  var modeKeys: Array[ModeKey] = null
 
   /** The format the mapper is reading from, set once per mapper from the subclass */
   val format: FactsetFormat
@@ -175,6 +175,8 @@ abstract class ChordFactsetMapper[K <: Writable] extends CombinableMapper[K, Byt
     ctx.thriftCache.pop(context.getConfiguration, ChordJob.Keys.FeatureIdLookup, featureIdLookup)
     entities = ChordJob.setupEntities(ctx.thriftCache, context.getConfiguration)
     emitter = MrContextEmitter(context)
+
+    modeKeys = ModeKey.fromDictionary(DictionaryCache.load(context.getConfiguration, ctx.thriftCache))
   }
 
   final override def setupSplit(context: MapperContext[K], split: InputSplit): Unit = {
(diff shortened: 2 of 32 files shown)


ivory-mr/src/main/scala/com/ambiata/ivory/mr/DictionaryCache.scala
```diff
@@ -0,0 +1,22 @@
+package com.ambiata.ivory.mr
+
+import com.ambiata.ivory.core.{Crash, Dictionary}
+import com.ambiata.ivory.core.thrift.{ThriftDictionary, DictionaryThriftConversion}
+import com.ambiata.poacher.mr.ThriftCache
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.mapreduce.Job
+
+// TODO Can/should I add a symmetric test for this - worth it?
```
markhibberd: Personally, I would do a round trip on the conversions and leave this to trusting thrift cache. If you did want to do it, maybe poacher should expose this: https://github.com/ambiata/poacher/blob/master/src/test/scala/com/ambiata/poacher/mr/ThriftCacheSpec.scala as a bit of kit to do this generally.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/ChordJob.scala
```diff
@@ -165,7 +165,8 @@ abstract class ChordFactsetMapper[K <: Writable] extends CombinableMapper[K, Byt
 
   val featureIdLookup = new FeatureIdLookup
 
-  var entities: Entities = null
+  var entities: Fact => Boolean = null
```
markhibberd: I am not sure why these bug me so much, I guess this one is ok though, it is total right?

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/ChordJob.scala
```diff
@@ -194,36 +197,15 @@ abstract class ChordFactsetMapper[K <: Writable] extends CombinableMapper[K, Byt
    * 2. chord.<version>.skip - number of facts skipped because they were in the future
    */
   override def map(key: K, value: BytesWritable, context: MapperContext[K]): Unit = {
-    ChordFactsetMapper.map(fact, converter, key, value, priority, kout, vout, emitter, okCounter, skipCounter, dropCounter, serializer,
-      featureIdLookup, entities)
+    converter.convert(fact, key, value)
```
markhibberd: I think this happens on the first line of the next map call as well. I wouldn't be adverse to not passing the converter down and doing it outside, but probably should be one or the other.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/ChordJob.scala
```diff
@@ -283,19 +267,23 @@ class ChordV2IncrementalMapper extends ChordIncrementalMapper[IntWritable] with
 
 object ChordIncrementalMapper {
```
markhibberd: Can we move/rename this object now that it is just the ChordMapper.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/ChordJob.scala
```diff
@@ -283,19 +267,23 @@ class ChordV2IncrementalMapper extends ChordIncrementalMapper[IntWritable] with
...
 
-  def map[K <: Writable](fact: MutableFact, key: K, value: BytesWritable, priority: Priority, kout: BytesWritable, vout: BytesWritable,
-                         emitter: Emitter[BytesWritable, BytesWritable], okCounter: Counter, skipCounter: Counter, dropCounter: Counter,
-                         serializer: ThriftSerialiser, featureIdLookup: FeatureIdLookup, entities: Entities, converter: MrFactConverter[K, BytesWritable]) {
+  def map[K <: Writable](fact: MutableFact, converter: MrFactConverter[K, BytesWritable], key: K, value: BytesWritable,
+                         priority: Priority, kout: BytesWritable, vout: BytesWritable,
+                         emitter: Emitter[BytesWritable, BytesWritable], okCounter: Counter, skipCounter: Counter,
+                         dropCounter: Counter, serializer: ThriftSerialiser, featureIdLookup: FeatureIdLookup,
+                         filter: Fact => Boolean, modeKeys: Array[ModeKey]) {
```
markhibberd: This is really out of control. Not sure what to do about it in this PR, but the awkward structure of this `map` on companion object is making things slowly get worse and worse, at some point we need to stop and fix it.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/SnapshotJob.scala
```diff
@@ -200,37 +204,15 @@ abstract class SnapshotFactsetMapper[K <: Writable] extends CombinableMapper[K,
    * 2. snapshot.<version>.skip - number of facts skipped because they were in the future
    */
   override def map(key: K, value: BytesWritable, context: MapperContext[K]): Unit = {
-    SnapshotFactsetMapper.map(fact, date, converter, key, value, priority, kout, vout, emitter,
-                              okCounter, skipCounter, dropCounter, serializer, featureIdLookup)
+    ChordIncrementalMapper.map(fact, converter, key, value, priority, kout, vout, emitter,
```
markhibberd: Yeh, that code really needs to be moved somewhere better.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/SnapshotJob.scala
```diff
@@ -200,37 +204,15 @@ abstract class SnapshotFactsetMapper[K <: Writable] extends CombinableMapper[K,
...
+      okCounter, skipCounter, dropCounter, serializer, featureIdLookup,
+      f => f.date <= date, modeKeys)
```
markhibberd: This seems non optimal, allocate the function on every map record.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/ChordJob.scala
markhibberd: The more I look at the use sites of this, the more it seems like it is making things sub-optimal. My personal view, is that this should be split into a series of smaller steps that each do a very small thing. Then the mapper should call out to assemble the right set of steps directly inline, rather than provide hole in the middle filter function etc...

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/mode/ModeKey.scala
```diff
@@ -0,0 +1,134 @@
...
+        val b = new Array[Byte](8)
+        v =>
+          writeLong(b, java.lang.Double.doubleToLongBits(v.getD), 0)
+          b
+      case DateEncoding =>
+        val b = new Array[Byte](4)
+        v =>
+          ByteWriter.writeInt(b, v.getDate, 0)
+          b
+    }
```
markhibberd: As discussed yesterday, I think this probably needs to change a bit (assume this is the PR comment refers to), but to answer your question, this shouldn't be using thrift.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/mode/ModeReducer.scala
```diff
@@ -56,3 +60,20 @@ object ModeReducerSet extends ModeReducer {
...
+  type X = Array[Byte]
+
+  def seed: Array[Byte] =
+    Array.emptyByteArray
+
+  def step(state: Array[Byte], f: Fact): Array[Byte] =
+    key.get(f)
+
+  def accept(state: Array[Byte], f: Fact): Boolean = {
+    // TODO We're having to look up this value twice for every fact - should we switch to single function (and introduce MutableTuple)?
```
markhibberd: Sorry, I couldn't see where this is actually called (just looking at the diffs, will have better look later), is it possible just to always call step and then pass the "stepped" state in here instead of this. Not sure, might solve the problem, but depends on how it is called I guess.

markhibberd:
Looks good, don't see any major issues once you sort out you thread the state & bytes through how you discussed.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/mode/ModeReducer.scala
charleso: > is it possible just to always call step and then pass the "stepped" state in here instead of this. 

Do you mean pass in two states in to this function?

I don't _think_ that will help/work. Especially in cases where the state is mutable it will override the previous value before you can compare that here. Happy to talk about this though - I think we should do _something_.

charleso:
Have rebased this on to `topic/mapper-cleanup-part-3` (Sorry for the noise - I really wish you could re-target PRs in Github). The `keyedSetNotEqual` spec is failing due to the shared-state bytes being clobbered. Will switch to `mundate-bytes` before finishing this up.

charleso:
@markhibberd This _should_ be ready now. There are two TODOs, one I'm fixing in Mundane now. The other is the previous question about `ModeReducerKeyedSet` and merging `accept`/`step`.

markhibberd:
Chat in the morning about todos, how far away are we from being able to add cli test?

charleso:
@markhibberd This should be ready now. Any final words? :)

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/mode/ModeReducer.scala
```diff
@@ -52,3 +59,46 @@ object ModeReducerSet extends ModeReducer {
...
 }
+
+class ModeReducerKeyedSet(key: ModeKey) extends ModeReducer {
+
+  import ModeReducerKeyedSet.State
+
+  type X = State
+
+  def seed: State =
+    State(Buffer.empty(255), Buffer.empty(255), DateTime.unsafeFromLong(-1))
```
charleso: Is there anything I can/should be doing with this? We're allocating new buffers on every entity.

markhibberd: Can't you just pull them out as vals?

charleso: > Can't you just pull them out as vals?

I can, just wanted to check that was OK.

markhibberd:
:+1:

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/mode/ModeReducer.scala
markhibberd: Seems like the right thing to do. The alternate is making them args, don't think it is particularly better or worse though.

PR #622, opened 2015-03-02
Introduce MrFactWriter and laws for MrFactConverter

charleso:
Let me know if `MrFactWriter` is a bad idea. If not, are there any "laws" I'm missing?


ivory-storage/src/main/scala/com/ambiata/ivory/storage/fact/MrFactConverter.scala
```diff
@@ -9,6 +9,10 @@ import org.apache.hadoop.io.{BytesWritable, IntWritable, NullWritable, Writable}
 sealed trait MrFactConverter[K <: Writable, V <: Writable] {
   def convert(fact: MutableFact, key: K, value: V): Unit
 }
+/** Currently only used for testing */
+sealed trait MrFactWriter[K <: Writable, V <: Writable] {
+  def write(fact: Fact, key: K, value: V): MrFactConverter[K, V]
+}
```
markhibberd: My only concern is that this interface _forces_ a copy even when we don't need one. Could be wrong though.

markhibberd:
I don't really get where/how this will be used (particularly given the other question), but I don't think there is problem with anything yet.

PR #623, opened 2015-03-02
Make FeatureIdIndex an actual value class

charleso:
@markhibberd I suspect you'll want me to rename the `apply` function.

/cc @raronson

### DIFF
--- a/ivory-core/src/main/scala/com/ambiata/ivory/core/FeatureIdMappings.scala
+++ b/ivory-core/src/main/scala/com/ambiata/ivory/core/FeatureIdMappings.scala
@@ -2,7 +2,17 @@ package com.ambiata.ivory.core
 
 import scalaz._, Scalaz._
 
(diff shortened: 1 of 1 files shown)


raronson:
:+1:

ivory-core/src/main/scala/com/ambiata/ivory/core/FeatureIdMappings.scala
```diff
@@ -2,7 +2,13 @@ package com.ambiata.ivory.core
 
 import scalaz._, Scalaz._
 
-case class FeatureIdIndex(int: Int) extends AnyVal
+class FeatureIdIndex(val int: Int) extends AnyVal
```
markhibberd: We have inconsistent use of toString on AnyVal's at the moment, I think we should probably be adding one, can be convinced otherwise if someone has a strong argument.

markhibberd:
Fine, other than toString comment.

PR #624, opened 2015-03-03
New poacher with fix when moving to a path whose parent doesn't exist

raronson:
Currently if you run snapshot with an output path whose parent doesn't exist, it will fail. The fix is in poacher where `mv` now creates the parent dir, see ambiata/poacher#30


charleso:
:+1:

charleso:
Actually do we need to update `notion`?

markhibberd:
Should bump in notion first, but otherwise :+1:

PR #625, opened 2015-03-03
Introduce FeatureIdIndexOption for dealing with FeatureIdLookup

charleso:

PR #626, opened 2015-03-03
Inline Chord/Snapshot map companions

charleso:
This obviously includes #625.


ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/ChordJob.scala
```diff
@@ -284,23 +281,12 @@ class ChordV2IncrementalMapper extends ChordIncrementalMapper[IntWritable] with
 
 object ChordIncrementalMapper {
```
charleso: Whoops. Need to remove these too.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/ChordJob.scala
```diff
@@ -195,36 +195,24 @@ abstract class ChordFactsetMapper[K <: Writable] extends CombinableMapper[K, Byt
...
+    if (!entities.keep(fact))
       skipCounter.count(1)
     else {
-      okCounter.count(1)
-      SnapshotWritable.KeyState.set(fact, priority, kout, featureId)
-      val bytes = deserializer.toBytes(fact)
+      SnapshotWritable.KeyState.set(fact, priority, kout, featureId.get)
+      val bytes = serializer.toBytes(fact)
       vout.set(bytes, 0, bytes.length)
       emitter.emit(kout, vout)
```
charleso: Not sure how much more to go with this serialisation block.

I was thinking about adding a `serializer.writeBytes(fact, bytes, 0)` function in `ThriftSerializer`. And/or something that coms out of `mundane-bytes`.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/ChordJob.scala
```diff
@@ -195,36 +195,24 @@ abstract class ChordFactsetMapper[K <: Writable] extends CombinableMapper[K, Byt
...
-          dropCounter: Counter, deserializer: ThriftSerialiser, featureIdLookup: FeatureIdLookup,
-          entities: Entities) {
-    val name = fact.featureId.toString
-    val featureId = featureIdLookup.getIds.get(name)
-    if (featureId == null)
+    val featureId = FeatureIdIndexOption.lookup(fact, featureIdLookup)
+    if (featureId.isEmpty)
       dropCounter.count(1)
-    else if (!entities.keep(fact))
+    if (!entities.keep(fact))
```
markhibberd: Missing else?

ivory-storage/src/main/scala/com/ambiata/ivory/storage/repository/RecreateFactsetJob.scala
```diff
@@ -109,14 +114,17 @@ class RecreateFactsetMapper extends Mapper[NullWritable, BytesWritable, BytesWri
   override def map(key: NullWritable, value: BytesWritable, context: MapperContext): Unit = {
     serializer.fromBytesViewUnsafe(thrift, value.getBytes, 0, value.getLength)
     val fact = createFact(thrift)
-    val k = featureIdLookup.ids.get(fact.featureId.toString).toInt
-
-    FactsetWritable.set(fact, kout, k)
+    val featureId = FeatureIdIndexOption.lookup(fact, featureIdLookup)
+    if (featureId.isEmpty)
+      dropCounter.count(1)
```
markhibberd: Mmmm, not sure about this, I assume it use to barf? Not sure if we should be dropping data here, kind of need to use the string feature id instead of the int one.

markhibberd: It looks like maybe the old behaviour was non-obvious but would of actually worked.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/ChordJob.scala
```diff
@@ -275,35 +263,25 @@ abstract class ChordIncrementalMapper[K <: Writable] extends CombinableMapper[K,
...
+    else if (!entities.keep(fact))
       skipCounter.count(1)
     else {
-      okCounter.count(1)
-      SnapshotWritable.KeyState.set(fact, priority, kout, featureIdLookup.getIds.get(fact.featureId.toString))
+      SnapshotWritable.KeyState.set(fact, Priority.Max, kout, featureId.get)
       val bytes = serializer.toBytes(fact)
       vout.set(bytes, 0, bytes.length)
       emitter.emit(kout, vout)
+      okCounter.count(1)
```
markhibberd: I assume this is a step in the right direction and there will be more work to de-dupe this afterwards?

charleso: > I assume this is a step in the right direction and there will be more work to de-dupe this afterwards?

As discussed. See last two commits. Not 100% sold on `SnapshotWritable.writeAndEmit`.

charleso:
@markhibberd @raronson ping?

raronson:
As per discussion IRL, im fine with this as long as the mapper code is tested through other specs.

The only slight concern is that the counters are not tested now.

markhibberd:
:+1:

PR #627, opened 2015-03-03
Use FeatureIdIndexOption in RecreateFactsetJob

charleso:
Isolating this change to continue discussion about missing featureId behaviour.

https://github.com/ambiata/ivory/pull/626#discussion_r25660540

### DIFF
--- a/ivory-storage/src/main/scala/com/ambiata/ivory/storage/repository/RecreateFactsetJob.scala
+++ b/ivory-storage/src/main/scala/com/ambiata/ivory/storage/repository/RecreateFactsetJob.scala
@@ -14,6 +14,7 @@ import com.ambiata.ivory.storage.task.{FactsetWritable, FactsetJob}
 import com.ambiata.ivory.storage.partition._
 import com.ambiata.mundane.control._
 import com.ambiata.mundane.io.{BytesQuantity, FilePath}
+import com.ambiata.poacher.hdfs.Hdfs
 import com.ambiata.poacher.mr._
 import org.apache.hadoop.fs.Path
 import org.apache.hadoop.io._
@@ -21,6 +22,7 @@ import org.apache.hadoop.mapreduce._
 import org.apache.hadoop.mapreduce.lib.input.{FileInputFormat, SequenceFileInputFormat}
 
 import scalaz.{Reducer => _, _}
+import scala.collection.JavaConverters._
 
 /**
  * This is a hand-coded MR job to read facts from factsets as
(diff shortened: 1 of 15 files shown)


ivory-storage/src/main/scala/com/ambiata/ivory/storage/repository/RecreateFactsetJob.scala
```diff
@@ -109,14 +110,17 @@ class RecreateFactsetMapper extends Mapper[NullWritable, BytesWritable, BytesWri
   override def map(key: NullWritable, value: BytesWritable, context: MapperContext): Unit = {
     serializer.fromBytesViewUnsafe(thrift, value.getBytes, 0, value.getLength)
     val fact = createFact(thrift)
-    val k = featureIdLookup.ids.get(fact.featureId.toString).toInt
+    val featureId = FeatureIdIndexOption.lookup(fact, featureIdLookup)
+    if (featureId.isEmpty)
+      Crash.error(Crash.Invariant, s"Unknown feature '${fact.featureId}' for entity '${fact.entity}'")
```
charleso: Yeah, so this did used to crash, but sounds like it shouldn't and just use the string instead?

markhibberd: I don't think this matches the current behaviour, unless I am misreading, and I don't think it is correct, it should be copying across all the facts.

charleso:
I have added a fix that uses `FeatureIdIndex(0)` and modified the test. It's currently failing because rename a `Factset` with 100% unknown features will result in 0 reducers. What is the expected behaviour in this case? /cc @raronson

ivory-storage/src/main/scala/com/ambiata/ivory/storage/repository/RecreateFactsetJob.scala
```diff
@@ -109,9 +110,10 @@ class RecreateFactsetMapper extends Mapper[NullWritable, BytesWritable, BytesWri
   override def map(key: NullWritable, value: BytesWritable, context: MapperContext): Unit = {
     serializer.fromBytesViewUnsafe(thrift, value.getBytes, 0, value.getLength)
     val fact = createFact(thrift)
-    val k = featureIdLookup.ids.get(fact.featureId.toString).toInt
+    val featureId = FeatureIdIndexOption.lookup(fact, featureIdLookup)
 
-    FactsetWritable.set(fact, kout, k)
+    // If we don't know about this feature any more just default to 0 - don't fail
+    FactsetWritable.set(fact, kout, if (featureId.isDefined) featureId.get else FeatureIdIndex(0))
```
markhibberd: Is 0 the starting index by default? If so, could we pick something larger, maybe `max(featureId) + 1` if it is possible. The main goal here is to return to a close as possible to perfect state - if there are features not currently in the dictionary I wouldn't wanting messing up compression of attribute 0.

markhibberd:
Besides the default index, I am good with this. In the 0 case, it should fall back to some dumb size calculation (1 reducer for ever 1GB of data maybe)

ivory-storage/src/main/scala/com/ambiata/ivory/storage/task/FactsPartitioner.scala
```diff
@@ -38,9 +38,13 @@ trait BaseFactsPartitioner[A] extends Partitioner[A, BytesWritable] with Configu
   def getConf: Configuration =
     _conf
 
-  def getPartition(k: A, v: BytesWritable, partitions: Int): Int =
-    FeatureReducerOffset.getReducer(lookup.reducers.get(getFeatureId(k)), getHash(k)) % partitions
+  def getPartition(k: A, v: BytesWritable, partitions: Int): Int = {
+    val offset = lookup.reducers.get(getFeatureId(k).int)
+    // If we don't know about this feature partition across all reducers
+    if (offset == null) getHash(k) % partitions
```
charleso: @markhibberd This is related to your comment about `max + 1` this morning. I don't think this is right. Should it be `featureId % partitions` instead, and does that mean we might get some nasty skew?

nhibberd:
@charleso bump

markhibberd:
@charleso Bump. I don't know context on this - if it is just tidy up, should we just close given how stale it is - or was this to address some issue?

charleso:
@markhibberd Yeah, basically tidy up, no issue.

PR #628, opened 2015-03-05
Introduce models for snapshot/chord for testing

charleso:
Hopefully this is not a step backwards. I would definitely argue the `SnapshotsSpec` has at least improved. `ChordArbitraries` is still bad, but I think we need to "model" squash too before that's really possible.

### DIFF
--- a/ivory-core/src/test/scala/com/ambiata/ivory/core/arbitraries/FactsWithStatePriority.scala
+++ b/ivory-core/src/test/scala/com/ambiata/ivory/core/arbitraries/FactsWithStatePriority.scala
@@ -1,0 +1,27 @@
+package com.ambiata.ivory.core.arbitraries
+
+import com.ambiata.disorder._
+import com.ambiata.ivory.core._
+import com.ambiata.ivory.core.arbitraries.Arbitraries._
+import org.scalacheck.Arbitrary, Arbitrary._
+
+case class FactsWithStatePriority(values: Map[DateTime, NaturalIntSmall]) {
+
+  def factsets(template: Fact): List[List[Fact]] = {
+    val max = values.values.map(_.value).max
+    (0 until max)
+      .map(i => values.filter(_._2.value > i).map(x => x._1 -> i).toList).toList
+      .map(_.map(x => {
+        val value = Value.unique(template.value, x._2)
+        Fact.newFactWithNamespace(template.entity, template.namespace, template.feature, x._1.date, x._1.time, value)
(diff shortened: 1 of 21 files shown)


ivory-core/src/test/scala/com/ambiata/ivory/core/arbitraries/FactsWithStatePriority.scala
```diff
@@ -0,0 +1,24 @@
...
+import com.ambiata.ivory.core._
+import com.ambiata.ivory.core.arbitraries.Arbitraries._
+import org.scalacheck.Arbitrary, Arbitrary._
+
+case class FactsWithStatePriority(values: Map[DateTime, NaturalIntSmall]) {
+
+  def factsets(template: Fact): List[List[Fact]] = {
+    val max = values.values.map(_.value).max
+    (0 until max).map(i => values.filter(_._2.value > i).map(x => x._1 -> i).toList).toList
+      // TODO Make value unique
```
markhibberd: Now or later?

ivory-operation/src/test/scala/com/ambiata/ivory/operation/model/ChordModel.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  def secondary: Order[S] =
+    SnapshotModel.secondary
+
+  def reduce(key: K, facts: NonEmptyList[Fact], conf: ChordModelConf): List[Fact] =
+    reduceWithDates(facts, conf.dictionary.byConcrete.sources(key._2), conf.entities.entities.get(key._1).toList.map(Date.unsafeFromInt))
+
+  def reduceWithDates(facts: NonEmptyList[Fact], dictionary: ConcreteGroup, dates: List[Date]): List[Fact] = {
+    (Date.minValue :: dates.reverse).sliding(2).toList.flatMap {
+      case List(d1, d2) =>
```
markhibberd: You can't filter a Nel, what happens if you filter out all the elements?

charleso: > You can't filter a Nel, what happens if you filter out all the elements?

Doh. Sorry, wasn't thinking.

markhibberd:
:japanese_castle: Looks good, can't see any issues/missing tests.

charleso:
Have added `Value.unique`, will merge when the tests are green.

PR #630, opened 2015-03-09
Introduce MutableOption

charleso:
@markhibberd This OK?

### DIFF
--- a/ivory-mr/src/main/scala/com/ambiata/ivory/mr/MutableOption.scala
+++ b/ivory-mr/src/main/scala/com/ambiata/ivory/mr/MutableOption.scala
@@ -1,0 +1,90 @@
+package com.ambiata.ivory.mr
+
+import com.ambiata.ivory.core.Crash
+
+import scalaz._, Scalaz._
+
+/**
+ * <pre>
+ *                  uuuuuuu
+ *               uu$$$$$$$$$$$uu
+ *           uu$$$$$$$$$$$$$$$$$uu
+ *          u$$$$$$$$$$$$$$$$$$$$$u
+ *         u$$$$$$$$$$$$$$$$$$$$$$$u
+ *        u$$$$$$$$$$$$$$$$$$$$$$$$$u
+ *        u$$$$$$$$$$$$$$$$$$$$$$$$$u
+ *        u$$$$$$"   "$$$"   "$$$$$$u
+ *        "$$$$"      u$u       $$$$"
+ *         $$$u       u$u       u$$$
+ *         $$$u      u$$$u      u$$$
+ *          "$$$$uu$$$   $$$uu$$$$"
+ *           "$$$$$$$"   "$$$$$$$"
+ *             u$$$$$$$u$$$$$$$u
+ *              u$"$"$"$"$"$"$u
+ *   uuu        $$u$ $ $ $ $u$$       uuu
+ *  u$$$$        $$$$$u$u$u$$$       u$$$$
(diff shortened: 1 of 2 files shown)


ivory-mr/src/main/scala/com/ambiata/ivory/mr/MutableOption.scala
```diff
@@ -0,0 +1,60 @@
...
+ *   u$$$uuu$$$$$$$$$uu ""$$$$$$$$$$$uuu$$$
+ *   $$$$$$$$$$""""           ""$$$$$$$$$$$"
+ *    "$$$$$"                      ""$$$$""
+ *      $$$"                         $$$$"
+ * </pre>
+ *
+ * This is to represent an optional value in cases where we wish to avoid the cost of an extra allocation.
+ * Please use with extreme care!
+ */
+class MutableOption[A](private var value: A, var isSet: Boolean) {
```
charleso: (Or is `OptionMutable` better?)

ivory-mr/src/main/scala/com/ambiata/ivory/mr/MutableOption.scala
```diff
@@ -0,0 +1,60 @@
...
+  def get: A = {
+    if (!isSet)
+      Crash.error(Crash.DataIntegrity, "The value is currently unset, please check `set` first")
+    value
+  }
+}
+
+object MutableOption {
+
+  def none[A](v: A): MutableOption[A] =
```
charleso: Is there a way to avoid this somehow?

markhibberd: You can covariant hack it. But I don't think it is necessary, but you probably want a `setNone` method so you can tweak an existing value to be none.

charleso: > but you probably want a setNone method so you can tweak an existing value to be none.

Whoops. Yes I do.

ivory-mr/src/main/scala/com/ambiata/ivory/mr/MutableOption.scala
markhibberd: Personally not adverse to just making it unicode skull and cross bones. But this is fine.

markhibberd: `case class` or `toString`/`equals` implementations. Prefer case class though, I don't think there is a good reason not to IMO.

ivory-mr/src/main/scala/com/ambiata/ivory/mr/MutableOption.scala
```diff
@@ -0,0 +1,60 @@
...
+ * Please use with extreme care!
+ */
+class MutableOption[A](private var value: A, var isSet: Boolean) {
+
+  def get: A = {
+    if (!isSet)
+      Crash.error(Crash.DataIntegrity, "The value is currently unset, please check `set` first")
+    value
+  }
+}
```
markhibberd: A `toOption` method would be a nice addition, probably end up useful for testing.

markhibberd:
:+1:

ivory-mr/src/main/scala/com/ambiata/ivory/mr/MutableOption.scala
charleso: > A toOption method would be a nice addition, probably end up useful for testing.

Ha. I had that and then removed it. Happy to put it back.

PR #631, opened 2015-03-09
Combine step/accept for ModeReducer using MutableOption

charleso:


ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/mode/ModeReducer.scala
```diff
@@ -11,8 +12,8 @@ trait ModeReducer {
   /** My kingdom for decent rank-2 types */
   type X
   def seed: X
-  def step(state: X, fact: Fact): X
-  def accept(state: X, fact: Fact): Boolean
+  /** Return the next state if the step is valid, or none otherwise */
+  def step(state: X, option: MutableOption[X], fact: Fact): MutableOption[X]
```
charleso: Should this be renamed to `stepOption` or something?

ivory-operation/src/test/scala/com/ambiata/ivory/operation/extraction/mode/ModeReducerSpec.scala
```diff
@@ -43,14 +44,8 @@ Construction
   def laws(mr: ModeReducer): Prop = new Properties("ModeReducer laws") {
     property("seed is always equal") =
         mr.seed ?= mr.seed
-    property("accept is consistent") = forAll((f: Fact) =>
-      mr.accept(mr.seed, f) ?= mr.accept(mr.seed, f)
-    )
     property("step is consistent") = forAll((f: Fact) =>
-      mr.step(mr.seed, f) ?= mr.step(mr.seed, f)
-    )
-    property("step and accept is consistent") = forAll((f: Fact) =>
```
charleso: Is there a new law I can introduce now?

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/mode/ModeReducer.scala
markhibberd: I wouldn't. Think it is fine.

ivory-operation/src/test/scala/com/ambiata/ivory/operation/extraction/mode/ModeReducerSpec.scala
markhibberd: Not that I can see from this, kind of supersedes.

markhibberd:
:palm_tree:

PR #632, opened 2015-03-10
Extra keyed_set to support compound keys

charleso:


ivory-core/src/main/thrift/core.thrift
```diff
@@ -142,6 +142,7 @@ enum ThriftDictionaryMode {
 union ThriftDictionaryModeV2 {
     1: ThriftDictionaryMode mode,
     2: string keyedSet
+    3: list<string> keyedSetMulti
```
markhibberd: Mmm... I think we should possibly bumping the metadata version with these changes - really we want things to not understand this to fail right? Need to think about it a bit more...

ivory-core/src/test/scala/com/ambiata/ivory/core/gen/GenDictionary.scala
```diff
@@ -10,7 +10,10 @@ import scalaz.scalacheck.ScalaCheckBinding._
 object GenDictionary {
   def mode: Gen[Mode] =
     // By generating a KeyedSet we force the encoding to be a struct, so let's generate less of them
-    Gen.frequency(9 -> modeImplemented, 1 -> GenString.sensible.map(Mode.KeyedSet))
+    Gen.frequency(
+      9 -> modeImplemented,
+      1 -> GenPlus.listOfSized(1, 3, GenString.sensible).map(Mode.KeyedSet)
+    )
 
   // Remove when KeyedSet is supported fully
```
markhibberd: Can this comment be removed now?

ivory-core/src/main/thrift/core.thrift
charleso: > really we want things to not understand this to fail right?

Yep, good point.

markhibberd:
Looks pretty good. Good test for implementation that we can make this sort of change without too much hassle.

ivory-core/src/main/thrift/core.thrift
```diff
@@ -1,5 +1,13 @@
 namespace java com.ambiata.ivory.core.thrift
 
+/**
+ * ====================================================================================================
+ *
+ * NOTE: Please update the Ivory metadata version for _any_ non-forward compatible change to this file.
```
charleso: Warning added

ivory-core/src/test/scala/com/ambiata/ivory/core/gen/GenDictionary.scala
charleso: > Can this comment be removed now?

The function should have been removed before in the previous PR. Done now.

PR #634, opened 2015-03-10
Bump mundane and replace Buffer allocate with grow

charleso:


charleso:
Merging. This change should, at least in theory, be a no-brainer.

PR #637, opened 2015-03-11
Introduce dictionary-metadata cli test

charleso:
Still need to figure out a way to deal with the exact commit that results in a version bump.


markhibberd:
:1234:

PR #638, opened 2015-03-11
Extract Value to separate file

charleso:


ivory-core/src/main/scala/com/ambiata/ivory/core/Value.scala
```diff
@@ -0,0 +1,248 @@
...
+      case (vp: PrimitiveValue, EncodingPrim(ep)) => validateEncodingPrim(vp, ep)
+      case (vp: PrimitiveValue, _)              => fail
+      case (StructValue(vs), EncodingStruct(es)) => validateStruct(vs, es)
+      case (StructValue(_),  _)                 => fail
+      case (ListValue(v), EncodingList(l))      => v.foldMap(validateEncodingSub(_, l.encoding))
+      case (ListValue(_),    _)                 => fail
+    }
+  }
+
+  def toThrift(value: Value): ThriftFactValue =
```
charleso: Not sure if I should create a sub `value` package will separate classes for these as well?

charleso: > Not sure if I should create a sub value package will separate classes for these as well?

Let me rephrase that. I'd like to do something to keep this file from ballooning out (eg. json). Is a sub-package the preferred way?

charleso:
@markhibberd Ping.

markhibberd:
Change looks good, but I am strongly adverse to sub-packages.

PR #640, opened 2015-03-14
Replace scopt with pirate

charleso:
The first commit leaves around `IvoryCmd` to reduce the scope of changes. The final commit removes `IvoryCmd` to just leave `pirate.Command`. In that commit the `IvoryCmd.repository` usage is quite sucky. There's almost certainly 1000 better ways to handle it.


ivory-cli/src/main/scala/com/ambiata/ivory/cli/IvoryCmd.scala
```diff
@@ -1,82 +1,28 @@
...
-  def diagnostic(repository: Repository, flags: IvoryFlags): RIO[Unit] =
+
+  def diagnostic(repository: Repository): RIO[Unit] =
     RIO.safe(System.err.println(
       s"""================================================================================
          |
          |Ivory:
          |  Version:             ${IvoryVersion.get.version}
          |  Path:                ${repository.root.show}
-         |  Planning Strategy:   ${flags.plan.render}
```
markhibberd: Is this forced by something?

charleso: > Is this forced by something?

No, just to extract a separate `Parse[RIO[IvoryFlags]]` which does it's own logging and is only used by a small subset of commands. Not a fan?

ivory-cli/src/main/scala/com/ambiata/ivory/cli/IvoryCmd.scala
```diff
@@ -88,53 +34,42 @@ object IvoryCmd {
...
-    var repoArg: Option[String] = None
-    var strategy: Option[StrategyFlag] = None
-    parser.opt[String]('r', "repository") action { (x, c) => repoArg = Some(x); c} text
-      "Path to an ivory repository, defaults to environment variable IVORY_REPOSITORY if set"
-    parser.opt[StrategyFlag]("plan-strategy") action { (x, c) => strategy = Some(x); c} optional() text
-      "Run with the specified plan strategy, one of: pessimistic - minimal IO, best answer, higher memory; " +
-      "conservative - higher IO, best answer, lower memory; optimistic - higher IO, good answer, quicker."
-    new IvoryCmd[A](parser, initial, IvoryRunner(config => c =>
+  /** Should _only_ be called by upgrade - everything else related to a repository should call [[repository]] */
+  def repositoryBypassVersionCheck: Parse[IvoryConfiguration => RIO[Repository]] = {
```
markhibberd: Side thought - do we have any test that verify `--help` and `--version` work on an out of date repository, I figure they don't go through this code at all, but would be nice to make sure.

charleso: > Side thought - do we have any test that verify --help and --version work on an out of date repository, I figure they don't go through this code at all, but would be nice to make sure.

We don't and it wouldn't, but happy to add.

ivory-cli/src/main/scala/com/ambiata/ivory/cli/IvoryCmd.scala
markhibberd: Not really. I _really_ don't want half a dozen little pieces of logging. I have been slowly but painfully removing them - to step back towards it is a bit of a no-no for me.

ivory-cli/src/main/scala/com/ambiata/ivory/cli/PirateReaders.scala
```diff
@@ -0,0 +1,53 @@
...
+import org.joda.time.{LocalDate, DateTimeZone}
+
+import pirate._
+
+object PirateReaders {
+  implicit val CharRead: Read[Char] =
+    Read.tryRead(str => {
+      val chars = str.toCharArray
+      chars.length match {
+        case 0 => throw new IllegalArgumentException(s"'${str}' can not be empty!")
```
markhibberd: Wat?

ivory-cli/src/main/scala/com/ambiata/ivory/cli/PirateReaders.scala
```diff
@@ -0,0 +1,53 @@
...
+      case None =>
+        throw new IllegalArgumentException(s"'${s}' is not a valid snapshot-id.")
+      case Some(s) =>
+        s
+    }, "SNAPSHOT_ID")
+
+  implicit val StrategyFlagRead: Read[StrategyFlag] =
+    Read.tryRead(s => StrategyFlag.fromString(s) match {
+      case None =>
+        throw new IllegalArgumentException(s"'${s}' is not a valid strategy.")
```
markhibberd: If this is something in Pirate forcing this, it needs to change before we pull into ivory - this is seriously not cool.

charleso: > If this is something in Pirate forcing this, it needs to change before we pull into ivory - this is seriously not cool.

Not to worry, this was just laziness from me to keep the code as it was. Can easily use the `Read` constructor.

markhibberd:
Thanks - looks great, just in no particular order, some combination of lazy question/comment:
 - We should fix the Read if that is ok.
 - Does this remove help/version it was happening implicitly through the stupid scopt banner thing - wouldn't want to lost that as a fair few scripts call it I think.
 - No cli test changes, so it is pretty much 100% compatible? I assume there will be a few quirks right at the edge - but if it is mostly the same that is pretty impressive. 
 - We should do an audit of the interfaces before we doc them. Once this is in, will be a heap easier to make them not horrible I think.

charleso:
> We should fix the Read if that is ok.

Happy to. See `pirate` for a change that will make that easier.

> Does this remove help/version it was happening implicitly through the stupid scopt banner thing

It removes the version unless you call `--version` explicitly. Would you like me to add it to the command description?

markhibberd:
No, that is fine as long as you can call it.

charleso:
Now using disjunction for `Read` - much nicer.

All the cli tests should be passing as well. I stuff up the argument ordering of input/output path strings in one command.

ivory-cli/src/main/scala/com/ambiata/ivory/cli/PirateReaders.scala
```diff
@@ -0,0 +1,40 @@
...
+import com.ambiata.mundane.io.{Bytes, BytesQuantity}
+
+import org.joda.time.{LocalDate, DateTimeZone}
+
+import pirate._
+
+import scalaz._, Scalaz._
+
+object PirateReaders {
+  implicit val CharRead: Read[Char] =
```
markhibberd: This should be in pirate. Can you put it on an issue so I don't forget.

ivory-cli/src/main/scala/com/ambiata/ivory/cli/chord.scala
```diff
@@ -1,45 +1,49 @@
...
 
-    help("help") text "shows this usage text"
-    opt[String]('c', "entities") action { (x, c) => c.copy(entities = x) }  required() text "Path to file containing entity/date pairs (eid|yyyy-MM-dd)."
-    opt[Unit]("no-snapshot")     action { (x, c) => c.copy(takeSnapshot = false) }     text "Do not take a new snapshot, just any existing."
-    opt[Int]("sample-rate") action { (x, c) => c.copy(squash = c.squash.copy(profileSampleRate = x)) } text
-      "Every X number of facts will be sampled when calculating virtual results. Defaults to 1,000,000. " +
-        "WARNING: Decreasing this number will degrade performance."
-  })(c => f => c.copy(formats = f(c.formats)))
+  , ( flag[String](both('c', "entities"), description("Path to file containing entity/date pairs (eid|yyyy-MM-dd)."))
+  |@| switch(long("no-snapshot"), description("Do not take a new snapshot, just any existing."))
```
markhibberd: This looks wrong. Logic flipped?

ivory-cli/src/main/scala/com/ambiata/ivory/cli/PirateReaders.scala
charleso: > This should be in pirate. 

Whoops, it already is.

ivory-cli/src/main/scala/com/ambiata/ivory/cli/chord.scala
charleso: > This looks wrong. Logic flipped?

Good pick up

charleso: > This looks wrong. Logic flipped?

I already know the answer - but worth adding a specific chord cli test for this?

markhibberd: I don't know if we spit out enough metadata to do it - we should really provide more info about how many jobs run etc...? I guess if we could kind of scrape the snapshot directory, not very civilised. Will leave it up to you - if you can find a decent way to test it, probably worth it (given this bug).

charleso:
Thanks to everyone who made this possible!

Programming in the Large

PR #644, opened 2015-03-16
Use a cache for the ivy repository

etorreborre:
See the documentation on this feature in the [sbt user guide](http://www.scala-sbt.org/0.13/docs/Travis-CI-with-sbt.html)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+# Use container-based infrastructure
+sudo: false
 language: scala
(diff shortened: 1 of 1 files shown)


markhibberd:
LGTM just want to wait until master is fixed so we can test this build first.

charleso:
Ignore one of the broken builds (flakey test, will look in to), it looks like it's working (or at least not breaking).

PR #645, opened 2015-03-16
Fix ci.guide script and publish to ambiata-oss

charleso:


etorreborre:
:+1:

bin/ci.guide
```diff
@@ -9,4 +9,4 @@ make build
 ./dist/build/guide/guide rebuild
 
 VERSION=$(date '+%Y%m%d%H%M%S')-$(git log --pretty=format:%h -n 1)
-aws s3 --region=ap-southeast-2 cp --recursive _site/ s3://ambiata-dist/ivory-guide/${VERSION}/
+aws s3 --region=ap-southeast-2 cp --recursive _site/ s3://ambiata-oss/com.ambiata/ivory-guide/${VERSION}/
```
markhibberd: If we added an extra sync to `ivory-guide/latest` would that help? So we can see can have a url to always check it for now.

markhibberd:
:shaved_ice:

PR #646, opened 2015-03-16
Add -D to pirate flags for proper parsing

charleso:
Not sure if this is what you had in mind @markhibberd.


markhibberd:
:+1: Assuming it all hangs together.

PR #647, opened 2015-03-20
Ensure all json fields are consumed (stops fields which aren't in the di...

FrontierOtr:
...ctionary from getting silently dropped)


markhibberd:
Spec before commit. But :gun:

charleso:
:+1:

ivory-core/src/test/scala/com/ambiata/ivory/core/ValueSpec.scala
```diff
@@ -47,6 +48,12 @@ class ValueSpec extends Specification with ScalaCheck { def is = s2"""
     Value.json.decodeJson(e.enc, Value.json.toJson(e.value).nospaces).toEither ==== Right(e.value)
   )
 
+  def jsonUnknown = {
+    val encoding = EncodingStruct(StructEncoding(Map("a" -> StructEncodedValue(IntEncoding, false))))
+    val json = """{"a":1,"b":2}"""
+    Value.json.decodeJson(encoding, json).toEither must beLeft
```
charleso: As discussed IRL it might be nice if this was property-based, but it's up to you...

PR #651, opened 2015-04-02
Switch to mkdocs for guide

charleso:


bin/ci.guide
```diff
@@ -1,24 +1,11 @@
...
-fi
+cd $(dirname $0)/../guide
 
 make build
 
-./dist/build/guide/guide rebuild
-
 VERSION=$(date '+%Y%m%d%H%M%S')-$(git log --pretty=format:%h -n 1)
-aws s3 --region=ap-southeast-2 cp --recursive _site/ s3://ambiata-oss/com.ambiata/ivory-guide/${VERSION}/
+aws s3 --region=ap-southeast-2 cp --recursive site/ s3://ambiata-oss/com.ambiata/ivory-guide/${VERSION}/
```
charleso: Not sure whether we should keep publishing these? Does seem like there's much harm.

guide/Makefile
```diff
@@ -1,28 +1,20 @@
...
+${MKDOCS}:
+	curl https://pypi.python.org/packages/source/m/mkdocs/${MKDOCS}.tar.gz | tar xz
 
-${DEPS}: ${SANDBOX} $(wildcard *.cabal)
-	cabal install -j --only-dependencies --enable-tests
-	cabal configure --enable-tests ${CABAL_FLAGS}
-	touch $@
+bin/mkdocs: ${MKDOCS}
+	cd ${MKDOCS} && python setup.py build
+	cd ${MKDOCS} && python setup.py install --user --install-scripts=../bin
```
charleso: Yeah I got lazy with `--user`. If there's a better way let me know.

charleso:
Interestingly, on the master branch we (already) get this:

http://ivory.readthedocs.org/en/latest/

I'm a little worried about how it's going to find the `mkdocs.yml` file unless it's in the root.

markhibberd:
@charleso We can move the current docs around to suit if we need to. This seems like a decent shortcut though.

charleso:
@markhibberd Looking at the code [here](https://github.com/rtfd/readthedocs.org/blob/32e7d496b99efaa1a9ce6e455e124c0793b363dc/readthedocs/doc_builder/backends/mkdocs.py) I think we at least need the `mkdocs.yml' file in the root. I've made a fixup commit that I think will work, but not sure if it's too invasive in to the root of the ivory dir.

markhibberd:
Given we are just trying to bootstrap things, this doesn't seem to bad, what do you think about using this as the first cut?

charleso:
@markhibberd Yeah I'm fine with this, except perhaps the `Makefile` being in the root. I might see if I can put it in the `guide` dir and still have everything work and then merge if that's OK?

markhibberd:
:1234:

PR #653, opened 2015-04-07
Return an empty array if there are no features to lookup 

etorreborre:
in `FeatureLookups.sparseMapToArray`.

### DIFF
--- a/ivory-storage/src/main/scala/com/ambiata/ivory/storage/lookup/FeatureLookups.scala
+++ b/ivory-storage/src/main/scala/com/ambiata/ivory/storage/lookup/FeatureLookups.scala
@@ -3,6 +3,7 @@ package com.ambiata.ivory.storage.lookup
 import com.ambiata.ivory.core._
 import com.ambiata.ivory.lookup._
 import scala.collection.JavaConverters._
+import scalaz._, Scalaz._
 
 object FeatureLookups {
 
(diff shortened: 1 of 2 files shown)


ivory-storage/src/main/scala/com/ambiata/ivory/storage/lookup/FeatureLookups.scala
```diff
@@ -53,11 +53,14 @@ object FeatureLookups {
     new EntityFilterLookup(features.map(_.toString).asJava, entities.asJava)
 
   def sparseMapToArray[A : scala.reflect.ClassTag](map: List[(Int, A)], default: A): Array[A] = {
-    val max = map.map(_._1).max
-    val array = Array.fill(max + 1)(default)
-    map.foreach {
-      case (i, a) => array(i) = a
+    if (map.isEmpty) Array()
+    else {
+      val max = map.map(_._1).max
```
charleso: There isn't a `maxOr` that we could use here is there? (Same for the `maxBy` in the test)

charleso:
:pray:

ivory-storage/src/main/scala/com/ambiata/ivory/storage/lookup/FeatureLookups.scala
etorreborre: Not that I know of in the standard lib. Maybe in Scalaz.

markhibberd: `maximum`/`maximumOf`/`maximumBy` if I remember correctly.

etorreborre:
I used `maximum` and `maximumBy` from Scalaz.

PR #649, opened 2015-03-23
initial draft

pmwtsn:

PR #654, opened 2015-04-08
initial commit

pmwtsn:
This supercedes PR #649.

### DIFF
--- a/guide/install.md
+++ b/guide/install.md
@@ -1,4 +1,71 @@
-Install
-=======
+---
+title: Install Ivory
+---
 
-- TODO
+Set up these prerequisites:
+
+## Java
+
+Java must be installed, and JAVA_HOME must be set.
+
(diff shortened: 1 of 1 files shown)


markhibberd:
@pmwtsn For reference: http://ivory.readthedocs.org/en/latest/install/

pmwtsn:
@markhibberd - thanks Mark. I'll fix the title.

Also, @blever noted that the curl command towards the end points to an old version... any idea what that should be now?

markhibberd:
@pmwtsn I will get it updated.

pmwtsn:
@markhibberd cheers

charleso:
:+1:

guide/install.md
```diff
@@ -1,4 +1,71 @@
-Install
-=======
+---
+title: Install Ivory
```
charleso: @pmwtsn FYI This should just be the normal markdown title (with the equal signs). We're not using hackyll any more, and this is now displayed incorrectly. Not a big deal, but just something to watch out for as you go.

PR #656, opened 2015-04-13
begin overview,dictionary, data-prep

pmwtsn:
Early drafts for:
* Overview
* Dictionary
* Data preparation

Assets folder for graphics.


guide/index.md
```diff
@@ -2,5 +2,7 @@ Ivory Guide
 ===========
 
 ### Table of Contents
-
+- [Overview](overview)
```
charleso: You may also want to add these to the `mkdocs.yml` file. Given that it controls the layout on the left, I wonder whether we want/need this index TOC at all.

guide/ingest/preparing-data.md
```diff
@@ -1,3 +1,35 @@
...
+A fact has these components:
+
+&lt;Entity&gt; &lt;Attribute&gt; &lt;Value&gt; &lt;Time&gt;
+
+That is, a fact represents the _value_ of an _attribute_ associated with an _entity_, which is known to be valid at some point in _time_.
+
+Examples of facts are:
+
+<table>
+    <tr><td>*Entity*</td><td>*Attribute*</td><td>*Value*</td><td>*Time*</td></tr>
```
charleso: You can do [markdown tables](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#tables), which has the benefit of supporting markdown in each table element (which the html version won't do). Also FYI this table isn't being rendered when you "view' this markdown file.

guide/dictionary.md
```diff
@@ -0,0 +1,31 @@
+# The Ivory dictionary
```
charleso: You'll probably want to add this to the `mkdocs.yml` file, currently it's not referenced by anything

guide/overview.md
```diff
@@ -0,0 +1,18 @@
+# Overview of Ivory
```
charleso: Same here - `mkdocs.yml`

guide/ingest/preparing-data.md
```diff
@@ -1,3 +1,35 @@
...
+
+The attributes referenced in facts must be declared in the Ivory repository's dictionary. The dictionary lists all known attributes and the metadata associated with them.
+
+##Transposing data
+
+You can use various tools to transpose data into a format that Ivory can ingest:
+
+* SQL
+* Python scripts
+* [Apache Thrift](https://thrift.apache.org/)
```
charleso: Thrift isn't really a tool for doing the transposition - it's one of the data formats.

guide/ingest/preparing-data.md
```diff
@@ -1,3 +1,35 @@
...
+
+Ivory can ingest facts in these formats:
+
+* Text - human readable, and can include JSON structs, but is not recommended for production due to performance reasons.
```
charleso: Should we link to `ingest/text-eavt.md` and `ingest/thrift-eavt.md` here?

guide/dictionary.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Entity attributes
+An attribute is identified in the dictionary by a namespace and name.  
+
+For example:
+
+<table>
+    <tr><td>*Namespace*</td><td>*Name*</td><td>*Encoding*</td><td>*Description*</td><tr>
+    <tr><td>demographic</td><td>gender</td><td>string</td><td>The customer's gender</td><tr>
+    <tr><td>demographic</td><td>zipcode</td><td>string</td><td>The customer's zipcode</td><tr>
```
charleso: Should we be using the weather attributes here instead as the single "working example"? /notsure

guide/dictionary.md
```diff
@@ -0,0 +1,31 @@
+# The Ivory dictionary
+The dictionary lists all known attributes and the metadata associated with them. All attributes referenced in facts must be declared in the Ivory repository's dictionary.
```
blever: It may be better to call out upfront that the dictionary lists 2 types of attributes - those that are referenced by facts, and those that derived/virtual features.

guide/dictionary.md
```diff
@@ -0,0 +1,31 @@
...
+
+For example:
+
+<table>
+    <tr><td>*Namespace*</td><td>*Name*</td><td>*Encoding*</td><td>*Description*</td><tr>
+    <tr><td>demographic</td><td>gender</td><td>string</td><td>The customer's gender</td><tr>
+    <tr><td>demographic</td><td>zipcode</td><td>string</td><td>The customer's zipcode</td><tr>
+    <tr><td>account</td><td>mthly_spend</td><td>double</td><td>The customer's account spend in the last month</td><tr>
+
+Namespaces are used as a data partitioning mechanism internally. Generally, attributes that are related should be contained in the same namespace.
```
blever: ... and similarly, those that are not related should be placed in separate namespaces. The rough rule of thumb for defining "relatedness" is if you would typically want to access features/attributes together.

guide/dictionary.md
blever: Schema for dictionary is actually `<namespace>|<name>|<key-value pairs>`. The current list of supported keys are:

* `encoding`
* `description`
* `source`
* `expression`
* `window`
* `filter`

And their use is somewhat context sensitive, depending on what type of attribute it is - physical or virtual.

@charleso are there others?

guide/dictionary.md
blever: Maybe good to define them as "physical" vs "virtual" attributes?

etorreborre: I like "base"/"derived" but I think we already had this conversation, didn't we?

guide/dictionary.md
charleso: > charleso are there others?

`tombstone` and `mode`

guide/ingest/preparing-data.md
```diff
@@ -1,3 +1,35 @@
...
+* Binary - generated using Thrift, for example, and recommended for production.
```
blever: @charleso is this really the recommendation? I've heard that JSON is faster than thrift for example ...

guide/dictionary.md
blever: Further to the above (and below), I think it's probably worth viewing attributes like the following:

* "physical"
  * are ingested as "facts"
  * are extracted as _latest_ "features"
  * define `mode` such as `state`, `set`, `keyed-set`
  * define `encoding`
* virtual
  * are extracted as "features"
  * define `source`
  * define `expression`, `window`, `filter`, etc

guide/ingest/preparing-data.md
charleso: >  is this really the recommendation?  I've heard that JSON is faster than thrift for example ...

I can't speak for extract and the compression, but thrift is definitely recommended for ingest.

blever: Ignore that ^^^. That's not actually true.

guide/overview.md
blever: This may or may not be useful, but this is the text for the summary and abstract of an Ivory talk I will be giving in a month (no slides yet):

```
Improving feature engineering in the lab and production with Ivory
==================================================================

Description (65)
----------------

Ivory is a new open-source, Hadoop-based data store that focuses on changing the
way we approach the critical and time consuming activity of scalable feature
engineering. It both simplifies and adds rigour to data science pipelines, aiding
in their transition from the lab to production environments.


Abstract (3 - 6 paragraphs)
---------------------------

Feature engineering is a critical and time consuming activity in the development
and deployment of any modeling pipeline. It is also exacerbated as data science teams seek to
incorporate new data sources into their pipelines that are at a scale far larger than
previously employed. Furthermore, the transition to production environments is littered with
complexity as these pipelines are exposed to the dynamic, and fragile, world of ongoing data
feeds, data corrections, and evolving data models.

In this talk we will introduce Ivory, a new open-source, Hadoop-based data store that seeks to
address these challenges. Ivory is a scalable and extensible data store for storing facts and extracting
features. It is optimised specifically for the feature engineering stages of modeling pipelines,
simultaneously simplifying and adding rigour to them.

This session will walk through an example of how Ivory can be used in the typical data
scientist's workflow, and then how that extends to migrating pipelines into production. It will impart
all of the basic concepts of Ivory such as repositories, the dictionary, its fact-based data
model, and virtual features. It will also demonstrate the benefits of Ivory being an immutable
data store and the unique opportunities that creates.


Prerequisite knowledge
----------------------

Attendees should have a working knowledge of Hadoop and be familiar with the
high-level machine learning activities involved in building a typical modelling
pipeline, e.g. data sourcing, feature engineering, training, testing and scoring.
```

guide/ingest/preparing-data.md
blever: Other tools:

* R
* Hive, Pig
* Custom MapReduce jobs or MapReduce libraries such as Scalding, Crunch, Cascalog and Scoobi

guide/ingest/preparing-data.md
```diff
@@ -1,3 +1,35 @@
...
+    <tr><td>*Entity*</td><td>*Attribute*</td><td>*Value*</td><td>*Time*</td></tr>
+    <tr><td>cust_00678</td><td>gender</td><td>M</td><td>2011-03-17</td></tr>
+    <tr><td>acnt_1234</td><td>balance</td><td>342.17</td><td>2014-06-01</td></tr>
+    <tr><td>car_98732</td><td>make</td><td>Toyota</td><td>2012-09-25</td></tr>
+
+The attributes referenced in facts must be declared in the Ivory repository's dictionary. The dictionary lists all known attributes and the metadata associated with them.
+
+##Transposing data
+
+You can use various tools to transpose data into a format that Ivory can ingest:
```
blever: FWIW, we often call this process "Fact ETL".

PR #660, opened 2015-06-16
Don't allow ignored items at the end of a subexpression

HuwCampbell:


charleso:
Thanks @HuwCampbell. :+1:

PR #665, opened 2015-07-24
Fix output dictionary example: was comma, should be pipe

tranma:
Just fixing up an error in the README. The output dictionary shouldn't have a comma after the column number, as consulted with @HuwCampbell and @charleso


charleso:
:+1: Maybe synchronise with @raronson who is about to merge too...

tranma:
rebased. I didn't know we prefer rebase over merge, I've been doing it wrong in other projects!

HuwCampbell:
:pear:

PR #666, opened 2015-07-26
Set mandatory service role for perf emr cluster

charleso:
To fix "InstanceProfile is required for creating cluster" errors

/cc @raronson @markhibberd


markhibberd:
:+1:

PR #671, opened 2015-08-03
Fix data-regression cli test

raronson:
@charleso Is this count needed? doesn't the above diff cover it?


raronson:
> Is this count needed?

Just looked again and noticed this was comparing the count from the previous snapshot. Ill add it back, but using `wc -l` as i removed the fact-count app.

charleso:
If this builds :+1:

PR #672, opened 2015-08-04
Remove scoobi

raronson:
Finally

@markhibberd @charleso


markhibberd:
Awesome :fireworks: 

Is this going to be an issue? https://github.com/ambiata/poacher/search?utf8=%E2%9C%93&q=scoobi Is it worth considering no mr1 support going forward? If that is the issue?

raronson:
> Is this going to be an issue?

Not sure. Think we're only using the compatibility libs through `DistCache` in ivory?

> Is it worth considering no mr1 support going forward?

We could, but we might hit the compatibility issues with future hadoop versions.

If we move `ScoobiAction` into its own lib, we could remove scoobi (except for the compatibility libs) from poacher also.

ivory-storage/src/test/scala/com/ambiata/ivory/storage/repository/RepositoryBuilder.scala
```diff
@@ -74,10 +72,56 @@ object RepositoryBuilder {
...
+    val locationIO = LocationIO(conf, Clients.s3)
+    for {
+      files <- locationIO.list(location)
+      facts <- filterHiddenLocations(files).traverseU(location => SequenceFileFactReaders.readMutableFacts(location, locationIO))
+    } yield facts.flatten
+  }
+
+  /* This isn't ideal, need a better way */
+  def filterHiddenLocations(locations: List[Location]): List[Location] =
+    locations.filterNot(l => l.render.matches(""".*/[\._][^/]+$"""))
```
charleso: Yeah, this isn't ideal. I'm assuming `Location` isn't supposed to have a `name` function?

charleso:
:100: 

I love the smell of deleted code in the morning

ivory-operation/src/test/scala/com/ambiata/ivory/operation/extraction/output/GroupByEntityOutputSpec.scala
```diff
@@ -82,6 +86,16 @@ class GroupByEntityOutputSpec extends Specification with ThrownExpectations with
     }
   }.set(minTestsOk = 5, maxDiscardRatio = 10)
 
+  def readThrift[A <: ThriftLike : Manifest](location: Location, locationIO: LocationIO, newThrift: () => A): RIO[List[A]] = {
```
charleso: Seems quite generic - might be good to move to extract somewhere?

raronson: Yeah, good idea. ill add more convenience functions also like `readThriftKeys`, `readThriftValues`, etc

charleso:
:+1:

PR #673, opened 2015-08-07
Do not run squash during a snapshot if there are no outputs (Fixes #661)

### LINKED ISSUE #661: Squash should only run with an external snapshot, not internal

raronson:
@nhibberd @charleso


charleso:
:+1:

nhibberd:
:cactus:

PR #674, opened 2015-08-07
Use EMR_DefaultRole for service role

charleso:


ivory-performance/src/test/bin/emr-start.sh
```diff
@@ -3,7 +3,7 @@ export AWS_DEFAULT_REGION=ap-southeast-2
 aws configure set preview.emr true
 
 EMR_CLUSTER_JSON=$(aws emr create-cluster --ami-version 3.0.1 --no-auto-terminate --tags Name=IvoryPerf \
-  --service-role=test.ci \
+  --service-role=EMR_DefaultRole \
```
markhibberd: This doesn't look like the right thing to me. We shouldn't be using anything but out roles.

PR #678, opened 2015-08-26
Remove namespaced thrift - step1

raronson:
`NamespacedThriftFact` is currently being used to transport facts from mappers to reducers and as the output format for squash and chord jobs. This isn't ideal as the namespace is causing unnecessary IO when it could be found in a lookup table or in a partitioned directory.

This PR  is the first step to remove its use by:
  - Introducing a new fact called a `HydratedFact` which contains a `ThriftFact` and mutable namespace and date variables. Its a replacement of `MutableFact`.
  - Send minimum amount of data between mappers and reducers, eg. only `ThriftFact` and `Date`

After this, the next step will be to change the output format of squash and chord to be partitioned by namespace, and store the date in the key and `ThriftFact` in the value (like snapshot).

Apologies for the size of the PR, `NamespacedThriftFact` cuts deep.

@charleso @markhibberd


markhibberd:
@raronson The code all looks pretty straight forward. I think it would pay to set up a good production regression test to run all of these changes through. (i.e. take a snapshot from last version on a real repository and then take a new one from each of the PRs and make sure they don't change). After that I am :+1:

charleso:
:baby_bottle:

nhibberd:
@raronson bump - whats the status?

PR #679, opened 2015-08-27
Updated notion to remove the stats uploading to CloudWatch

etorreborre:
@markhibberd @charleso @nhibberd


ivory-storage/src/main/scala/com/ambiata/ivory/storage/sync/Sync.scala
```diff
@@ -73,4 +75,11 @@ object Sync {
     }
   } yield infos
 
+
+  /** run a distcopy and save statistics to S3 and CloudWatch */
+  def runDistCopy(mappings: Mappings, conf: DistCopyConfiguration): RIO[Unit] =
+    for {
+      stats <- DistCopyJob.run(mappings, conf)
+      _     <- stats.cata(ss => StatisticsMetrics.upload(ss.statisticsMetrics.addDimension(MetricDimension("application", "ivory")), "exsum"), RIO.ok(()))
```
nhibberd: Would this fail without cloudwatch permissions?

charleso: Seconded. I have to wonder whether there's a different way to we can publish/store the metrics and have them fed in to cloudwatch as a separate step? Or at least this would have to be optional. Might be good to discuss first.

etorreborre: @nhibberd maybe I can check the permissions first and warn if they are missing. Another option is to pass an option in the configuration to enable/disable the upload of stats.

@charleso the separate step is provided in the form of using `view` to download/upload the repo and deal with statistics (in that case `ivory` won't do any distcopy).

etorreborre: @nhibberd thinking about your question. Yes this should fail without permissions. Actually this should behave like it currently does on master where the upload code is in `notion`. Why would it be different?

nhibberd:
@etorreborre @raronson Is this ok to rebase and merge now?

etorreborre:
@nhibberd this looks ok to me. @raronson are you ok with a merge?

project/depend.scala
```diff
@@ -19,7 +19,7 @@ object depend {
   // NOTE: We have a copy of TDeserializer in core that needs to be kept in sync (or removed) when thrift is updated
   val thrift    = Seq("org.apache.thrift"    %  "libthrift"       % "0.9.1" excludeAll ExclusionRule(organization = "org.apache.httpcomponents"))
 
-  val sawsVersion = "1.2.1-20150721073540-3b8a713"
+  val sawsVersion = "1.2.1-20150827031911-183969b"
```
raronson: Something is a bit weird, I don't see `183969b` in the list of commits on https://github.com/ambiata/saws/commits/master

etorreborre: Actually we should drop the `saws` dependencies as it comes transitively with `notion`. I propose to use [`notion-d9374b9`](https://github.com/ambiata/notion/commit/d9374b9f93d59cc95ed090b348239c968491efac) which drags [`saws-105d82b`](https://github.com/ambiata/saws/commit/105d82be8d47730f8d19eb2083c5a21a7f1a229e).

raronson: Do the upload/download counters matter for this? if so, https://github.com/ambiata/notion/commit/78b38eedefbeacc4a16721d794ab452f59236cbb would probably be better.

etorreborre: I don't think so since we're not using them anymore in Ivory. But a more recent notion version works as well (see the fixup).

etorreborre:
@raronson let's merge?

raronson:
The branch build looks like its broken

```
04:18:32  Running jar tvf /mnt/jenkins/work/ivory-branches/ivory-cli/target/scala-2.11/ivory-cli-assembly-1.0.0-cdh5-20151030040921-2cb9e56.jar
04:18:34 java.lang.RuntimeException: Jar is too fat! Number of files in jar is 66686, needs to be less than 65536
```

etorreborre:
@raronson fun times... I'm going to have a look at that.

etorreborre:
@raronson one obvious culprit is mundane which depends on specs2 for production classes. This is actually just one part of the story since many useless jars are being brought in through transitive dependencies.

image

markhibberd:
@etorreborre For reference that test is explicitly to avoid bringing specs2 in which what breaks the limit. The practice around all of these code bases is to depend on individual saws/mundane projects so that mundane-testing isn't brought in (which should be the only thing that depends on specs).

etorreborre:
@markhibberd what do you think of putting all test classes (matchers and arbitraries) in the `test` directory of `mundane-testing` and require a `"test"->"test"` dependency for projects wanting to use those matchers? One drawback is that the matchers are tested with a spec which would also be accessible from client jars.

etorreborre:
@raronson :+1: ?

raronson:
:+1:

PR #681, opened 2015-08-28
Remove namespaced thrift step2

raronson:
This is the follow on from #678. Changes:

  * Output format of chord and squash are partitioned by namespace, and stored with the date as key and `ThriftFact` as value.
  * `GroupByEntityOutputJob` uses `ThriftFact` instead of `NamespacedThriftFact` when passing facts from mappers to reducers.
  * New partitioned emitter because squash reducer outputs to multiple namespaces.


ivory-core/src/main/scala/com/ambiata/ivory/core/Fact.scala
```diff
@@ -89,53 +89,6 @@ object Fact {
...
-
-    def value: Value =
-      Value.fromThrift(fact.getValue)
-
-    def toThrift: ThriftFact = fact
-
-    def toHydrated: HydratedFact =
-      HydratedFact(namespace, date, toThrift)
-
-    def toNamespacedThrift = this
```
charleso: You have a `TODO` in the other PR for removing this from the `trait`. It looks like the only place left is in `NamespacedFactWriter`, can/should that be removed?

ivory-storage/src/main/scala/com/ambiata/ivory/storage/fact/Namespaces.scala
```diff
@@ -37,10 +37,16 @@ object Namespaces {
...
-  def fromSnapshotPath(path: Path): Namespace = {
+  def fromPathUnsafe(path: Path): Namespace = {
     Namespace.nameFromStringDisjunction(path.getParent.getName) match {
       case scalaz.\/-(n) => n
-      case scalaz.-\/(e) => Crash.error(Crash.DataIntegrity, s"Can not parse snapshot namespace from path '${path}' - ${e}")
+      case scalaz.-\/(e) => Crash.error(Crash.DataIntegrity, s"Can not parse namespace from path '${path}' - ${e}")
     }
   }
+
+  def sizeNamespaces(root: Path): Hdfs[List[Sized[Namespace]]] =
```
charleso: Worth adding a separate test for?

charleso:
:baby: 

FYI the cli tests are failing, although looks like a map ordering problem

ivory-core/src/main/scala/com/ambiata/ivory/core/Fact.scala
raronson: thanks, i missed this, removing

ivory-storage/src/main/scala/com/ambiata/ivory/storage/fact/Namespaces.scala
raronson: actually, i think its a duplicate of `namespaceSizes`, so will remove

nhibberd:
@raronson status

PR #683, opened 2015-09-03
Add psv json cell sorting of output for comparison

raronson:
This is for when json output is sorted differently across java versions

@charleso @markhibberd


charleso:
Sweet. So easy. This python thing is pretty cool.

:baby:

domdere:
finally some modern technology around here

PR #688, opened 2015-09-16
Order squash facts by value with same entity/feature/time

raronson:
This fixes non-determinism in output of squash

@markhibberd @charleso


ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/squash/SquashReducerState.scala
```diff
@@ -18,21 +19,42 @@ object EntityIterator {
...
+    val prevFact = createMutableFact
     var value: A = initial
+
+    def update(): Unit = {
+      prevFact.setNspace(state.namespace.name)
+      prevFact.setYyyyMMdd(state.datetime.date.int)
+      prevFact.toThrift.setEntity(state.entity)
+      prevFact.toThrift.setAttribute(state.attribute)
+      if(state.datetime.time == Time(0)) prevFact.toThrift.unsetSeconds else prevFact.toThrift.setSeconds(state.datetime.time.seconds)
+      prevFactValues.sorted.foreach(tvalue => {
```
charleso: You might want to check the size. :(

https://github.com/scala/scala/blob/0b47dc2f28c997aed86d6f615da00f48913dd46c/src/library/scala/collection/SeqLike.scala#L636

charleso:
@raronson What's the plan with this?

raronson:
@charleso im going to add a switch to optionally turn it on because its only really needed for testing

markhibberd:
Can be re-opened if we really want to do it. Given it is a for testing-only thing i think I am against it, and it is no longer blocking other testing.

PR #693, opened 2015-10-19
Add ability to override number of reducers on group by entity job

raronson:
@markhibberd @charleso


ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/output/GroupByEntityOutput.scala
```diff
@@ -5,15 +5,19 @@ import com.ambiata.ivory.storage.lookup.ReducerSize
 import com.ambiata.mundane.control._
 import com.ambiata.mundane.io.MemoryConversions._
 
-import scalaz.effect.IO
+import scalaz._, Scalaz._, effect.IO, \&/._
 
 object GroupByEntityOutput {
+  val GroupByEntityNumReducers = "GroupByEntity.reducers"
```
markhibberd: If it is ok, can you prefix it with `ivory.`

charleso:
:+1:

raronson:
@charleso i pushed the property to the edge and passed it in

charleso:
:+1:

PR #694, opened 2015-10-23
Hadoop doesn't do system properties

raronson:
`ivory snapshot -Divory.extraction.GroupByEntity.reducers=xxx ...` doesn't set the system property `ivory.extraction.GroupByEntity.reducers`, but sets it in the hadoop configuration instead.

@charleso @markhibberd


markhibberd:
Is this a result of pirate switch? If so needs to be fixed, if not, need to
set as JAVA_OPTS how we set other jvm options. This should not be using
configuration.

raronson:
Found `HADOOP_OPTS` which looks like it sets the system properties as expected, closing

PR #695, opened 2015-11-05
Topic/time partition new

HuwCampbell:
Client request, not for production (so probably shouldn't be merged, just reviewed).


charleso:
CLI tests failing?

HuwCampbell:
Isn't that the regression test, where we need to commit another update (with the git hash) after the fact?

charleso:
> Isn't that the regression test, where we need to commit another update (with the git hash) after the fact?

All the tests are failing. I'm pretty sure that's not right. I believe you're thinking of the `dictionary-metadata` test, but that shouldn't affect everything else. Correct me if I'm wrong though.

ivory-core/src/main/scala/com/ambiata/ivory/core/MetadataVersion.scala
```diff
@@ -56,6 +60,7 @@ object MetadataVersion {
       case "v4" => V4
       case "v5" => V5
       case "v6" => V6
+      case "v7" => V7
       case x => Unknown(x)
     })
```
charleso: You haven't updated `latestVersion` below - that might be what's breaking things

ivory-operation/src/test/scala/com/ambiata/ivory/operation/extraction/reduction/TimePartitionReducerSpec.scala
```diff
@@ -0,0 +1,64 @@
...
+    r.update(Fact.newFact("","", "", Date.minValue, Time.unsafe(0), StructValue(Map("b" -> StringValue("2015-01-01T00:00:00.0000Z")))))
+    r.update(Fact.newFact("","", "", Date.minValue, Time.unsafe(0), StructValue(Map("b" -> StringValue("2015-01-01T02:01:20Z")))))
+    r.update(Fact.newFact("","", "", Date.minValue, Time.unsafe(0), StructValue(Map("b" -> StringValue("2015-01-01T04:01:20+10:00")))))
+    r.update(Fact.newFact("","", "", Date.minValue, Time.unsafe(0), StructValue(Map("b" -> StringValue("2015-01-01T05:01:20-01:00")))))
+    r.update(Fact.newFact("","", "", Date.minValue, Time.unsafe(0), StructValue(Map("b" -> StringValue("2014-01-01T23:59:59Z")))))
+    Value.fromThrift(r.save) ==== ListValue(List(1,0,1,0,1,1,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,1).map(x => LongValue(x.toLong)))
+  }
+
+  def laws =
+    ReducerUtil.reductionLaws(new CountReducer)
```
charleso: This isn't testing your `Reducer` is it?

HuwCampbell: Indeed.

ivory-core/src/main/scala/com/ambiata/ivory/core/MetadataVersion.scala
HuwCampbell: Done, ta.

HuwCampbell: It would be nice if Scala had Bounded, so latest was just maxBound.

ivory-core/src/main/scala/com/ambiata/ivory/core/MetadataVersion.scala
```diff
@@ -56,9 +60,10 @@ object MetadataVersion {
...
 
   def previousVersions = List(V0, V1, V2, V3, V4, V5)
```
charleso: Missing `V6` here :(

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/reduction/TimePartitionReducer.scala
```diff
@@ -0,0 +1,98 @@
...
+    case tv if tv.isSetS =>
+      try {
+        tv.getS match {
+          case pattern(y,m,d,h,mm,s,_,_) => for {
+              date <- Date.create(y.toShort,m.toByte,d.toByte)
+              time <- Time.create(h.toInt * 3600 + mm.toInt * 60 + s.toInt)
+            } yield (date.addTime(time))
+          case _ => tombstoneOut = true; None
+        }
+      } catch { case _ : Throwable => None}
```
charleso: Please don't catch `Throwable`, that includes things like `OutOfMemory` and other such things

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/squash/StateReduction.scala
```diff
@@ -39,6 +40,7 @@ class StateReduction(windowStart: Date, windowEnd: Date, reduction: Reduction) e
     // We're lucky (or unlucky) that thrift will recreate the value object each time and so it's safe (for now) to do this
     factPrev.getFact.setValue(fact.toThrift.getValue)
     factPrev.setYyyyMMdd(fact.toNamespacedThrift.getYyyyMMdd)
+    factPrev.getFact.setSeconds(fact.toNamespacedThrift.getFact.getSeconds)
```
charleso: It's all a little hazy now, but I recall that `seconds` is optional, so _if_ you're going to do this, you probably need to respect whether seconds is set or not. /cc @raronson

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/reduction/TimePartitionReducer.scala
HuwCampbell: Ok. I think this function is actually safe now, so I should be able to just remove the try catch.

ivory-operation/src/test/scala/com/ambiata/ivory/operation/extraction/reduction/TimePartitionReducerSpec.scala
```diff
@@ -0,0 +1,64 @@
...
+    r.update(Fact.newFact("","", "", Date.minValue, Time.unsafe(0), StructValue(Map("b" -> StringValue("2015-01-01T00:00:00.0000Z")))))
+    r.update(Fact.newFact("","", "", Date.minValue, Time.unsafe(0), StructValue(Map("b" -> StringValue("2015-01-01T02:01:20Z")))))
+    r.update(Fact.newFact("","", "", Date.minValue, Time.unsafe(0), StructValue(Map("b" -> StringValue("2015-01-01T04:01:20+10:00")))))
+    r.update(Fact.newFact("","", "", Date.minValue, Time.unsafe(0), StructValue(Map("b" -> StringValue("2015-01-01T05:01:20-01:00")))))
+    r.update(Fact.newFact("","", "", Date.minValue, Time.unsafe(0), StructValue(Map("b" -> StringValue("2014-01-01T23:59:59Z")))))
+    Value.fromThrift(r.save) ==== ListValue(List(1,0,1,0,1,1,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,1).map(x => LongValue(x.toLong)))
+  }
+
+  def laws =
+    ReducerUtil.reductionLaws(new TimePartitionReducer(None, DayOfWeek, () => new CountReducer))
```
charleso: Would prefer that `DayOfWeek` was arbitrary here and in the other properties (or is there a reason it can't be?)

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/squash/StateReduction.scala
HuwCampbell: I might just remove this. Previously, when we used this reducer for a client off a branch, I allowed times to be used.

raronson: Yes, `seconds` is optional and there needs to be an extra check if its set if your going to use the thrift object directly

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/reduction/TimePartitionReducer.scala
```diff
@@ -0,0 +1,98 @@
...
+  def update(fact: Fact): Unit = {
+    (dtFrom match {
+      case None             => Some(fact.datetime)
+      // The field is potentially optional, so wrap it in an option.
+      case Some(field)      => Option(fact.toThrift.getValue.getStructSparse.getV.get(field)).flatMap(retrieveDateTime)
+    }) map { inspectedDate =>
+      d match {
+        case WeekEndWeekDay =>
+          // Magic +2 ahead is because 1600-03-01 was a Wednesday, but also day 0, so to bring it up to Monday is 0, Sunday is 6, we add 2.
+          if ((DateTimeUtil.toDays(inspectedDate.date) + 2) % 7 < 5) reducers(0).update(fact) else reducers(1).update(fact)
```
charleso: This array indexing is more unsafe than it needs to be. You can/should just create a closure at the stop based on `d` that creates the hard-coded number of `Array`s that it needs. If not a closure a `trait` would work too.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/reduction/TimePartitionReducer.scala
```diff
@@ -0,0 +1,98 @@
...
+      } catch { case _ : Throwable => None}
+    case _                  => Crash.error(Crash.DataIntegrity, s"Time partition reducer has hit a value with the wrong type.")
+  }
+
+  def update(fact: Fact): Unit = {
+    (dtFrom match {
+      case None             => Some(fact.datetime)
+      // The field is potentially optional, so wrap it in an option.
+      case Some(field)      => Option(fact.toThrift.getValue.getStructSparse.getV.get(field)).flatMap(retrieveDateTime)
+    }) map { inspectedDate =>
```
charleso: `map` here is a lie, probably better make this `foreach`?

The other option is to make two different `Reducer`s, once for the structs, and one for the top-level values. That's what I've done in other places from memory. I'd actually prefer that although I realise it means extracting a few things out to share between the two.

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/reduction/TimePartitionReducer.scala
```diff
@@ -0,0 +1,98 @@
...
+
+  val reducers: Array[Reduction] = d match {
+    case DayOfWeek      => Array.fill(7)(f())
+    case WeekEndWeekDay => Array.fill(2)(f())
+    case HourOfDay      => Array.fill(24)(f())
+    case TimeOfDay      => Array.fill(4)(f())
+  }
+
+  var tombstoneOut = false
+  val pattern = new Regex("^([0-9]{4})-([0-9]{2})-([0-9]{2})T([0-9]{2}):([0-9]{2}):([0-9]{2})(.[0-9]+)?(Z|[+-][0-9]{2}:[0-9]{2})")
```
charleso: Do we _really_ need this regex? Can you use any of the `DateTime` functions we already have?

charleso: Is there anything in [Dates.scala](https://github.com/ambiata/ivory/blob/master/ivory-core/src/main/scala/com/ambiata/ivory/core/Dates.scala) that helps?

ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/squash/StateReduction.scala
HuwCampbell: I'm happy to either fix (add a check) or remove this. What do you prefer?

charleso: > I'm happy to either fix (add a check) or remove this. What do you prefer?

If you don't need it for your reduction then definitely remove. :)

ivory-operation/src/test/scala/com/ambiata/ivory/operation/extraction/reduction/TimePartitionReducerSpec.scala
```diff
@@ -0,0 +1,64 @@
...
+  def tombstoneOut = {
+    val r = new TimePartitionReducer(Some("b"), DayOfWeek, () => new CountReducer)
+    r.update(Fact.newFact("","", "", Date.minValue, Time.unsafe(0), StructValue(Map("b" -> StringValue("bad")))))
+    Value.fromThrift(r.save) ==== TombstoneValue
+  }
+
+  def parsing = {
+    val r = new TimePartitionReducer(Some("b"), HourOfDay, () => new CountReducer)
+    r.update(Fact.newFact("","", "", Date.minValue, Time.unsafe(0), StructValue(Map("b" -> StringValue("2015-01-01T00:00:00.0000Z")))))
+    r.update(Fact.newFact("","", "", Date.minValue, Time.unsafe(0), StructValue(Map("b" -> StringValue("2015-01-01T02:01:20Z")))))
```
charleso: Not mentioned IRL - if this is about testing parse, would be calling a `parse` function directly rather than going through `update()`

charleso:
Missing `Gen` instances for your expression

charleso:
As discussed, this is pretty evil. :+1: but let's hope that `icicle` saves us soon...

HuwCampbell:
Added generators, found an issue with the validation as well which is good.

HuwCampbell:
Squashed and merging.

PR #697, opened 2015-11-06
Set a proper cdh4 version

etorreborre:
@raronson I thought that notion versions were going hand in hand `cdh5` and `cdh4`. This fix uses the latest notion version for `cdh5` before the `yarn/mr1` change and the latest `cdh4` version.


markhibberd:
@etorreborre @charleso @phickey Can you coordinate on Monday to get the Yarn + IA changes into here. I would prefer if we didn't hold those back.

charleso:
Have discussed with @etorreborre IRL, he should be updating Ivory to the latest now

etorreborre:
@charleso please look at 4d768aa

charleso:
Thanks @etorreborre. Remember to fix (and rename) the relevant builds.

I'm assuming this change will also need a test in production due to @phickey's change (but I'm not sure).

charleso:
@etorreborre I think given @phickey change to the AWS jars this is going to need a production test before we role it out. Do we use `ivory` latest everywhere, or do we have to bump a version. If the former, we need to test this pretty soon.

PR #699, opened 2015-11-09
Removed scoobi classes from the assembly jar

etorreborre:
To avoid breaking the file number limit. @raronson


etorreborre:
After discussion, I am removing the `jarTest` check instead.

PR #701, opened 2015-11-11
Sort psv lines in cli tests

charleso:
/cc @raronson 

Not sure why this just started happening, but 3 of the cli tests are failing on most (but not all) of the cli builds.


raronson:
:+1:  after green build

PR #704, opened 2015-11-25
Increase chord reducers

raronson:
There's a lot more computation and less data in chord squash reducers then in snapshot ones, so we should increase the number of reducers

@markhibberd @charleso


charleso:
Surface :+1:

markhibberd:
:+1:

PR #706, opened 2015-11-26
Fix group by entity test

raronson:
@charleso @markhibberd This fixes the test that broke + i fixed some code which didn't look right in `GroupByEntityOutputJob`


charleso:
:+1:

Issue #662, opened 2015-07-06
Adding a mandatory field to a struct did not fail the dictionary import


nhibberd:
Ivory `1.0.0-cdh5-20150303004251-8d81cdd`

raronson:
Confirmed not an issue, but bad error message which was fixed through #707

PR #707, opened 2015-12-08
Fix non optional struct field error message

raronson:
@nhibberd @charleso 

I tested #662 and got a failure, but it was a misleading error message. This fixes that message


raronson:
Got :+1: IRL from @nhibberd, merging

charleso:
Thanks @raronson :+1:

PR #709, opened 2015-12-08
Fail when ivory repo doesn't exist

raronson:
@markhibberd @charleso


charleso:
Maybe a cli test as well.

:+1:

PR #710, opened 2015-12-08
Fix SparseOutputSpec test

raronson:
@charleso @nhibberd Some values are empty strings, and `EavtParsers.splitLine` calls `String.split` without `-1`.


charleso:
:+1:

markhibberd:
It is doing an additional check that it is exactly 4 fields. I can't tell from the context of this, is it EAVT or is it somthing that happens to be 4 fields. If it is EAVT, why is time empty I guess?

raronson:
> is it EAVT or is it somthing that happens to be 4 fields

Its EAV (without time) but namespace and attribute are in two columns. Its the sparse output job - https://github.com/ambiata/ivory/blob/master/ivory-operation/src/main/scala/com/ambiata/ivory/operation/extraction/output/SparseOutputJob.scala#L136-L142

markhibberd:
:+1:

PR #711, opened 2015-12-10
Fix bug where repo exists check only works on repositories that are multiple levels deep

raronson:
@markhibberd @charleso When i tested the original code, it was on a repo under `/tmp/ivory` which worked, but for some reason it doesn't work when the repo is at `/ivory`. This should work as its used by `create-repository`


charleso:
:+1:

raronson:
tested and seems to work, merging

PR #713, opened 2015-12-14
Reduce amount of data in GroupByEntity mr job value

raronson:
@markhibberd @charleso


markhibberd:
That looks huge. Will have a better look after lunch, but seems like a very good thing (tm).

charleso:
Yeah, what mark says. :100: assuming all the tests pass.

markhibberd:
:+1: after :green_apple:

PR #714, opened 2015-12-23
Commit annotations

EduardSergeev:
Generic mechanism for annotating `ivory` commits with a set of key=value pairs.

First intended use is to store `napoleon`s commits corresponding to exported dictionaries.


ivory-cli/src/main/scala/com/ambiata/ivory/cli/IvoryCmd.scala
```diff
@@ -79,14 +79,15 @@ object IvoryCmd {
...
+      case MetadataVersion.V0 | MetadataVersion.V1 | MetadataVersion.V2 | MetadataVersion.V3 |
+           MetadataVersion.V4 | MetadataVersion.V5 | MetadataVersion.V6 | MetadataVersion.V7 =>
         RIO.failIO[Unit](s"""The version of the ivory repository you are trying to access has
                                 |meta-data in a form which is too old to be read, you need to run
                                 |run `ivory update` in order for this version of ivory to proceed.
                                 |
                                 |WARNING: If you run `ivory update` older ivory installs will no
                                 |longer be able to access the repository.""".stripMargin)
-      case MetadataVersion.V7 =>
+      case MetadataVersion.V8 =>
```
charleso: You'll need to change https://github.com/ambiata/ivory/blob/master/ivory-cli/src/test/bin/tests/data-regression/expected/metadata/config#L2 this as well to get the cli tests to pass

ivory-storage/src/main/scala/com/ambiata/ivory/storage/metadata/AnnotationTextStorage.scala
```diff
@@ -0,0 +1,73 @@
...
+    val key = Repository.annotation(id)
+    for {
+      exists <- repository.store.exists(key)
+      annotation <-
+      if (exists)
+        for {
+          lines <- repository.store.utf8.read(key)
+          result <- RIO.fromDisjunctionString[Annotation](Parse.decodeEither[Annotation](lines))
+        } yield result
+      else Annotation.empty.point[RIO]
```
charleso: I'm not sure this is right (but apologies if I've done something like this elsewhere). If you ask for an annotation that doesn't exist it should return `None`. One suggestion would be to make `annotationSetId` `Option[AnnotationId]` for legacy repositories.

EduardSergeev: Yes, I've just realised that it was wrong (when I tried to plug it into cli). Will make it identical to other *Storages' behaviour.

charleso:
Just bikeshedding with names, but how does "notes" sound rather than "annotations"?

ivory-cli/src/test/bin/tests/commits/test
```diff
@@ -18,3 +18,6 @@ commit_test 00000002
 
 $IVORY import-dictionary -p "${INPUT}/dictionary.psv"
 commit_test 00000003
+
+$IVORY update-annotation -a napoleon-commit=00001234
```
charleso: I might be tempted to have an `annotation` command with `update` as a sub-sub command. Thoughts?

Either way be good to have a `get-annotation` command for testing the value of `napoleon-commit` here

ivory-operation/src/main/scala/com/ambiata/ivory/operation/annotation/AnnotationManager.scala
```diff
@@ -0,0 +1,26 @@
+package com.ambiata.ivory.operation.annotation
+
+import com.ambiata.ivory.core.{CommitId, Annotation, Repository}
+import com.ambiata.ivory.storage.control._
+import com.ambiata.ivory.storage.metadata.{Metadata, AnnotationTextStorage}
+import com.ambiata.mundane.control.RIO
+
+
+object AnnotationManager {
```
charleso: We would normally just call this `Annotation` :)

EduardSergeev: It clashes with [this Annotation](https://github.com/ambiata/ivory/blob/topic/commit-annotation/ivory-core/src/main/scala/com/ambiata/ivory/core/Annotation.scala) then (inspiration for name came to me from [DictionaryImporter](https://github.com/ambiata/ivory/blob/master/ivory-operation/src/main/scala/com/ambiata/ivory/operation/ingestion/DictionaryImporter.scala)). Any other suggestion for name or how avoid the clash?

EduardSergeev:
> Just bikeshedding with names, but how does "notes" sound rather than "annotations"?

Sounds shorter :) I'd even prefer "tags" but I think this name didn't play well with some other "tags".

EduardSergeev:
@charleso @raronson Minimal but complete (for napoleon needs so far) implementation.

PS I kind of still like `annotation` vs `note`, but willing to change if other people prefer it (or something else).

markhibberd:
@EduardSergeev I am :+1: for annotation.

ivory-cli/src/main/scala/com/ambiata/ivory/cli/annotation/get.scala
```diff
@@ -0,0 +1,34 @@
...
+    |""".stripMargin)
+
+  , (     flag[CommitId](both('c', "commit"), empty)
+      |@| IvoryCmd.repository
+
+  )((commitId, repo) => IvoryRunner(configuration => repo(configuration).flatMap(repository =>
+
+      IvoryT.fromRIO(for {
+          annotation <- AnnotationManager.getAnnotation(repository, commitId)
+          result = annotation.map(Annotation.toJson(_)).getOrElse("")
```
charleso: I wonder if you should just return an empty List from this command?

charleso:
:+1:

ivory-cli/src/main/scala/com/ambiata/ivory/cli/annotation/get.scala
EduardSergeev: Yes, actually I was meant to ask people's opinion on how who we handle missing annotations. `"{}"`, `""` or some other indication? Empty JSON seems ok to me: simpler handling on client side too (do we want to distinguish the cases of empty annotation vs absent one?).

charleso: Yeah I don't know. Returning `{}` seems fine for now.

EduardSergeev:
Bump @markhibberd `<|>` @raronson `<|>` @nhibberd  OK to merge?

nhibberd:
Will defer to @markhibberd @raronson

ivory-cli/src/main/scala/com/ambiata/ivory/cli/IvoryCmd.scala
```diff
@@ -10,7 +10,7 @@ import java.util.UUID
 
 import org.apache.hadoop.fs.Path
 
-import pirate._, Pirate._
+  import pirate._, Pirate._
```
markhibberd: Accidental space?

ivory-storage/src/main/scala/com/ambiata/ivory/storage/metadata/CommitStorage.scala
```diff
@@ -9,10 +9,11 @@ import scalaz._, Scalaz._
 import com.ambiata.mundane.control._
 import com.ambiata.ivory.core._
 
-// 3 lines.
+// 4 lines.
 // The first line is dictionaryId: DictionaryId
 // The second line is featureStoreId: FeatureStoreId
 // The third line is an optional repositoryConfigId: RepositoryConfigId
+// The fourth line is an optional set of annotations : Map[String, Annotation]
```
markhibberd: Assuming here, repository-config is mandatory if annotation line is set?

ivory-cli/src/test/bin/tests/annotation/test
```diff
@@ -0,0 +1,27 @@
...
+$IVORY import-dictionary -p "${INPUT}/dictionary.psv"
+commit_test 00000003
+
+$IVORY annotation get -c 00000003 > "${TARGET}/empty.json"
+annotation_test empty.json
+
+$IVORY annotation update -a napoleon-commit=00001234
+commit_test 00000004
+
+$IVORY annotation get -c 00000004 > "${TARGET}/annotation.json"
```
markhibberd: Would be good to add an extra check to this, for the case where the next ingest happens, that maintains the annotation.

markhibberd:
:+1: After extra check added to cli (if that comment makes sense)

markhibberd:
@EduardSergeev Actually, can you grab @raronson and myself this afternoon to discuss this one. Want to make sure we all understand the flow.

ivory-storage/src/main/scala/com/ambiata/ivory/storage/metadata/CommitStorage.scala
EduardSergeev: Must be, being the fourth line (after the third one). Do you mean it should be expressed here more explicitly? Or is it possible for new repositories to not have "repository config" line?

PR #715, opened 2015-12-24
Fix intermittent failure of EntitiesSpec test

EduardSergeev:
Filter fact by entity before passing them to `keepBestFacts` since it seems to expect facts for a specific `entityId` only

Rare test failure which indicated the problem: https://ci.ambiata.com/job/ivory.branches/57/consoleFull


markhibberd:
Based on the prod code, this looks correct I think. @charleso should double check though

charleso:
:+1:

Looks like `keepBestFacts` isn't used any more since we went to the raw chord MR. We should probably just delete it.

EduardSergeev:
Deleted it as per @charleso suggestion. Tests passed locally, strange errors in ci..

PR #717, opened 2016-01-17
Use AMBIATA_IVY_OSS in ci.guide

EduardSergeev:


charleso:
:+1: 

Although we should remove this build - it's not going to be used any more (AFAIK)

PR #718, opened 2016-01-19
Fix failing GroupByEntityOutputSpec

raronson:
@charleso @markhibberd Sane issue as #710 but with `GroupByEntityOutputSpec`


charleso:
:+1:

ivory-operation/src/test/scala/com/ambiata/ivory/operation/extraction/output/GroupByEntityOutputSpec.scala
```diff
@@ -43,7 +43,7 @@ class GroupByEntityOutputSpec extends Specification with ThrownExpectations with
     RepositoryBuilder.using(createDense(fwd.facts, fwd.dictionary,
       GroupByEntityFormat.DenseText(Delimiter.Psv, "NA", TextEscaping.Delimited, TextFormat.Json)) {
       (_, file) =>
-        IvoryLocation.readLines(file).map(l => parse(fwd.dictionary, "NA", l.map(EavtParsers.splitLine('|', _))))
+        IvoryLocation.readLines(file).map(l => parse(fwd.dictionary, "NA", l.map(_.split("\\|", -1).toList)))
```
markhibberd: We should remove that from mundane, it is just broken I think.

markhibberd: Or at least make it clear what is happening somehow.

There used to be something in mundane that did the "right" thing for this case, is it gone?

markhibberd:
:+1: regardless of what we do with mundane.

Issue #719, opened 2016-01-20
Add "get latest" set of operations to CLI


EduardSergeev:
Should return the latest (current):
 - annotation: `ivory annotation get [-c commit_id]` - output annotation for the latest (or `-c commit_id`) ivory commit 
 - commit_id: `ivory commit_get` (`ivory commit get` ?) - output latest ivory commit_id
 - dictionary_id, factset_id...?

1st is definitely needed for napoleon integration. Others (which ones?) could be useful as well?

PR #721, opened 2016-01-20
Annotation get latest

EduardSergeev:
For annotations only from: https://github.com/ambiata/ivory/issues/719


ivory-operation/src/main/scala/com/ambiata/ivory/operation/annotation/AnnotationManager.scala
```diff
@@ -2,15 +2,16 @@ package com.ambiata.ivory.operation.annotation
...
 import com.ambiata.mundane.control.RIO
 import scalaz._, Scalaz._
 
 
 object AnnotationManager {
 
-  def getAnnotation(repository: Repository, commitId: CommitId): RIO[Option[Annotation]] = {
+  def getAnnotation(repository: Repository, commitIdOpt: Option[CommitId]): RIO[Option[Annotation]] = {
     RepositoryT.runWithRepo(repository, for {
+      commitId <- RepositoryT.fromRIO(repo => commitIdOpt.map(_.pure[RIO]).getOrElse(CommitStorage.latestIdOrFail(repo)))
```
charleso: I vaguely wonder whether it's worth moving this line to `CommitStorage` as a helper, and then forcing a non-optional `CommitId` for `getAnnotation`. Would be easier to test each function that way. /uptoyou

charleso:
:+1:

raronson:
:+1:

ivory-operation/src/main/scala/com/ambiata/ivory/operation/annotation/AnnotationManager.scala
EduardSergeev: Hm.. I believe we still need to pass `Option[CommitId]` to `getAnnotation` (how otherwise would we indicate that we just want the latest `CommitId`, not the specific one?). I've tried the factor it out into [byIdOrLatestOrFail](https://github.com/ambiata/ivory/blob/topic/get-latest/ivory-storage/src/main/scala/com/ambiata/ivory/storage/metadata/CommitStorage.scala#L40-L41) (terrible name?). Is it something you've had in mind?

PR #722, opened 2016-01-20
Fix dictionary-metadata test

EduardSergeev:
By updating IVORY_BASELINE for new metadata version


charleso:
:+1:

ivory-cli/src/test/bin/ivory-get
```diff
@@ -11,7 +11,8 @@ mkdir -p "$DIST"
 if [ ! -e "${DIST}/ivory-cli-${VERSION}" ]; then
   # Redirect stdout to stderr so we don't confuse consumers who rely on just the path being "returned"
   # Note that before 20140704013114-9062d18 we didn't have have a tar distribution, only the jar
-  aws s3 cp "s3://ambiata-oss/com.ambiata/ivory-cli_${SCALA_VERSION}/${VERSION}/dists/ivory_${SCALA_VERSION}.tgz" "${DIST}" 1>&2
+  TGZ_PATH=com.ambiata/ivory-cli_${SCALA_VERSION}/${VERSION}/dists/ivory_${SCALA_VERSION}.tgz
+  (aws s3 cp "s3://ambiata-oss/$TGZ_PATH" "${DIST}" || aws s3 cp "s3://ambiata-oss-v2/$TGZ_PATH" "${DIST}") 1>&2
```
EduardSergeev: @charleso I don't know how to script, so this is probably something which could be written better (suggestions?). But it does do the trick: we now have to deal with two different S3 repositories in tests.

charleso: Thanks Eduard. You might want to exit  if that second command fails - something like `(...) || exit $? 1>&2`

Issue #720, opened 2016-01-20
Add optional `annotation` parameter to ivory commands


EduardSergeev:
Instead of running `annotation update` as a separate add annotation as a part other ivory operation.

Affected commands:
- create-repository
- import-dictionary
- ingest
- snapshot

EduardSergeev:
I was wondering how this "inline annotation" should behave internally. To be specific, I see the following two scenarios:

 1. There is no difference between applying annotation "inline" or doing it as a separate `annotation update` command - the end result, including the internal Ivory data will be identical (i.e. first the main operation is applied, then the annotation is added as a separate commit);
 2. When annotation is applied "inline" with another ivory command both the command's changes and the annotation are recorded in the same commit (i.e. in the final state there will be `-1` less commits in the repository).

I am not sure which way is more in line with the general ivory design, so I need your opinion on this, @markhibberd, @raronson, @charleso.

In terms of implementation `1.` seems to be simple and i like the fact the there will be no difference internally regardless of how we apply annotation (one or two ivory commands). `2.` on the other hand will lead to less frequently incremented commits but will require more changes in codebase and semantically two ways of annotating will be slightly different (internally).

Thoughts? 

PS: The current [WIP](https://github.com/ambiata/ivory/pull/723) takes `2.` approach.

markhibberd:
@EduardSergeev Haven't looked at code yet, but 2 sounds closer to what I would expect - the benefit of the inline one is that it can be attached at the commit it is relevant to, means that the annotations on the operation commit aren't the old ones.

EduardSergeev:
@markhibberd Yes, good point. This way we effectively annotate ivory **operation** instead of adding annotation post-factum to the state of the repository after operation. No chance to get "stale" annotation after that indeed.

PR #723, opened 2016-01-20
Add optional annotations parameter to `import-dictionary`

EduardSergeev:
Implements "import-dictionary" part of https://github.com/ambiata/ivory/issues/720


ivory-operation/src/main/scala/com/ambiata/ivory/operation/ingestion/DictionaryImporter.scala
```diff
@@ -47,5 +50,5 @@ object DictionaryImporter {
   case object Update extends ImportType
   case object Override extends ImportType
 
-  case class ImportOpts(ty: ImportType, force: Boolean)
+  case class ImportOpts(ty: ImportType, force: Boolean, annotation: Option[Annotation])
```
charleso: Would we want/need multiple annotations?

charleso: Just ignore me...

charleso:
Looks good so far :+1:

ivory-operation/src/main/scala/com/ambiata/ivory/operation/ingestion/DictionaryImporter.scala
```diff
@@ -36,8 +38,9 @@ object DictionaryImporter {
       doImport = validation.isSuccess || importOpts.force
       dictIdIden <- if (doImport)
         for {
+          annIdO <- importOpts.annotation.traverseU(a => RepositoryT.runWithRepo(repository, AnnotationTextStorage.store(a)))
```
charleso: Should this be merging annotations?

EduardSergeev: Yes, if `--merge` flag is supplied. Still WIP, will add it before PR is fully ready for review.

EduardSergeev:
I am leaving here only "import-dictionary" part of the task.

Other commands will be extended in separate PR's if needed.

EduardSergeev:
Bump @charleso I believe this is it for `import-dictionary`
OK to merge?

ivory-cli/src/main/scala/com/ambiata/ivory/cli/annotation/Args.scala
```diff
@@ -2,15 +2,21 @@ package com.ambiata.ivory.cli.annotation
...
-import pirate._, Pirate._
+import com.ambiata.ivory.operation.annotation.AnnotationManager._
+import pirate._, pirate.Pirate._
 import scalaz._, Scalaz._
 
 object Args {
 
+  def annotationArgs: Parse[AnnotationOpts] =
+      ( annotationDef
+    |@| switch(long("merge"), description("Merge given annotation with existing one instead of replacing it"))
```
charleso: I'm wondering whether merge should be the default or not?

charleso:
:+1:

ivory-cli/src/main/scala/com/ambiata/ivory/cli/annotation/Args.scala
EduardSergeev: Now I am wondering too :) Previously I though that `replace` would be more sensible behaviour but now that you've mentioned it I am no longer sure. Do we have any other potential usages on annotations in mind? If there is only "napoleon-commit" key it doesn't matter but with other annotations in place `merge` may be more convenient default. @markhibberd to the rescue?

markhibberd: I think the fact that replace implies you know what all the other annotations are, means it isn't the best default. I have some ideas for other annotations, but I don't think that it matters, the chance that two independent jobs might accidentally clobber each others annotations if they don't know about each other, makes merge seem like the only reasonable default. (Replace may only be useful if you were doing housekeeping / maintenance and wanted to clear out old annotations or similar maybe)

EduardSergeev: Yes, makes total sense. Merging by default and `--replace` flag is added.

